PR #27, opened 2024-08-05
Use SCM for titiler-cmr version

maxrjones:
This PR adjusts the project metadata to dynamically determine version based on the git history. The benefit of this approach is that you can distinguish between a release/tag, builds from commits after a release, and builds from uncommitted changes. I find this information especially helpful for performance benchmarks because it allows easily associating results with a given version of the codebase.

For example, the result of `python -c 'from titiler.cmr import __version__ as version; print(version)'` on this branch is `0.1.dev15+gb763dd1.d20240805`. The specific formatting of the version string is [configurable](https://backend.pdm-project.org/metadata/#dynamic-project-version).


vincentsarago (approved):
really cool 👍

PR #30, opened 2024-10-14
add /statistics, /feature, and /bbox endpoints

### LINKED ISSUE #29: Add /statistics POST endpoint
Add ability to send a geojson and get a statistics report for a CMR query

hrodmn:
The `rasterio` backend worked out-of-the-box but the `xarray` backend required a few small tweaks to which parameters get passed to the reader. The logic for handling options conditionally for each backend is kind of a mess, I would like to clean that up somehow but I tested the `/statistics` endpoint on HLS data and MUR SST data and it works!

Right now you can pass `max_size/height/width` to the `rasterio` backend but those args will be ignored for `xarray` backend.

Since they are so similar, I added the `/bbox` and `/feature` endpoints, too.

I was having some trouble reproducing the exact test behavior from my local environment in GH actions so I switched everything to use `uv` for installation/running things in CI and it worked the first time.

Resolves #29


vincentsarago (approved):

hrodmn:
@vincentsarago I know you already approved but do you want to take another quick look since I added the `/feature` and `/bbox` endpoints? Thanks!

titiler/cmr/backend.py
```diff
@@ -73,7 +74,7 @@ class CMRBackend(BaseBackend):
 
     input: str = attr.ib("CMR", init=False)
 
-    _backend_name = "CMR"
+    _backend_name: str = attr.ib(default="CMR")
```
vincentsarago, line 77: why do we need this change?

vincentsarago (approved):

titiler/cmr/backend.py
hrodmn: I kept getting linter warnings before making this change. It is possible I have things set incorrectly in my environment, but I think the result is the same with either approach and now I don't get the warning!
image

PR #33, opened 2024-10-22
add /timeseries endpoints

hrodmn:
The line count looks scary but the big changes here include ~600 lines of code in the new `titiler/cmr/timeseries.py` file. The rest is in a notebook, a VCR cassette, and tests.

- Based on spec and discussion from https://github.com/NASA-IMPACT/veda-architecture/pull/503
- Add /timeseries/statistics and possibly a few other endpoints

The goal is to start refining the timeseries API spec by testing out a bunch of CMR collections and seeing what kind of parameters make sense for most cases.

To see some examples of the `/timeseries` endpoints in action and some descriptions of the API for specifying a timeseries, check out the [new timeseries notebook](https://github.com/developmentseed/titiler-cmr/blob/feature/timeseries/docs/src/examples/Timeseries_Example.ipynb)

I am really interested in getting some feedback on the specification of the timeseries parameters. It feels like there are a lot of knobs so maybe there are ways to simplify it.


titiler/cmr/timeseries.py
```diff
@@ -0,0 +1,588 @@
...
+
+        current = next_step
+
+        if current == end:
+            ranges.append((end,))
+
+    if not ranges:
+        return [(start, end)]
+
+    return ranges
```
hrodmn: All of this would eventually move to `titiler.extensions.timeseries` but I wanted to do the development all in one place to facilitate easier iteration during early stages of development.

titiler/cmr/timeseries.py
hrodmn, line 1: A few of the classes and models defined at the top of this module will eventually go to `titiler.extensions.timeseries` so they can be re-used by other applications.

titiler/cmr/timeseries.py
```diff
@@ -0,0 +1,612 @@
...
+            description="End datetime for timeseries request",
+        ),
+    ] = None
+    step: Annotated[
+        Optional[str],
+        Query(
+            description="Time step between timeseries intervals, expressed as [ISO 8601 duration](https://en.wikipedia.org/wiki/ISO_8601#Durations)"
+        ),
+    ] = None
+    step_idx: Annotated[
```
abarciauskas-bgse: Not sure I understand what this is used for based on just the description - is it used when exact=False and the `step_idx` item is chosen from items returned for a single range request?

titiler/cmr/timeseries.py
```diff
@@ -0,0 +1,612 @@
...
+        Optional[int],
+        Query(description="Optional (zero-indexed) index of the desired time step"),
+    ] = None
+    exact: Annotated[
+        Optional[bool],
+        Query(
+            description="If true, queries will be made for a point-in-time at each step. If false, queries will be made for the entire interval between steps"
```
abarciauskas-bgse, line 145: I think (will try to inspect more below) that when `exact=False`, this means statistics are returned from only one item within the interval (that is, not an average across multiple items in that interval). If that is the case, perhaps we should make it more clear in the description and documentation (e.g. the notebook).

titiler/cmr/timeseries.py
```diff
@@ -0,0 +1,612 @@
...
+            description="Optional list of comma-separated specific time points or time intervals to summarize over"
+        ),
+    ] = None
+
+
+timeseries_field_names = [field.name for field in fields(TimeseriesParams)]
+
+
+def parse_duration(duration: str) -> relativedelta:
+    """Parse ISO 8601 duration string to relativedelta."""
```
abarciauskas-bgse: This may be a bit more clear if it gave an example - e.g. "duration=P1W" returns a relativedelta datetime object which can be used to create valid step-wise datetimes.

titiler/cmr/timeseries.py
```diff
@@ -0,0 +1,612 @@
...
+                format="GIF",
+                save_all=True,
+                append_images=pngs[1:],
+                loop=0,
+                duration=1000 // fps,
+            )
+
+            gif_bytes.seek(0)
+
+            return StreamingResponse(gif_bytes, media_type=TimeseriesMediaType.gif)
```
abarciauskas-bgse, line 623: 🚀

titiler/cmr/timeseries.py
```diff
@@ -0,0 +1,612 @@
...
+            post_process=Depends(factory.process_dependency),
+            stats_params=Depends(factory.stats_dependency),
+            histogram_params=Depends(factory.histogram_dependency),
+            image_params=Depends(factory.img_part_dependency),
+        ):
+            urls = build_request_urls(
+                base_url=str(factory.url_for(request, "geojson_statistics")),
+                request=request,
+                param_list=query,
+            )
```
abarciauskas-bgse, line 426: Just to make sure I understand how this works: the timeseries params are generated as part of the dependency `timeseries_query_no_bbox` - which generates a query object that is list of objects each containing a concept id and datetime. Build request urls returns a list of urls, each of which is a statistics endpoint, and then in the line below a request is made to each of those statistics endpoints. Is that understanding correct?

abarciauskas-bgse (approved):
This all looks good to me, I just made comments about clarifying some of the function and dependency descriptions. I focused on the notebook and titiler/cmr/timeseries.py. If time allows, I would like to understand how the VCR and data/podaac... files are being used (presumably in the tests).

titiler/cmr/timeseries.py
hrodmn: I thought `step_idx` would be useful in the case where you have a timeseries that you are interested in, but for some reason you only want to fetch results for a single time step at a time.

You could make requests like this:
```python
# format the base request url
base_request_url = "/timeseries/statistics?concept_id=1234&start_datetime=2024-10-01T00:00:01Z&end_datetime=2024-10-10T00:00:01Z"

first_request = f"{base_request_url}&step_idx=1"

second_request = f"{base_request_url}&step_idx=2"
```

One possible use-case would be if we add a `/timeseries/tiles` family of endpoints where you could easily get an xyz tile URL for a point along a timeseries. Maybe would be useful in a map application since you would have to load a single tile layer at a time. The `/timeseries/tilejson.json` endpoint gives you the whole set so maybe `step_idx` is not really necessary.

I'm fine with leaving this parameter out but it might be useful someday.

titiler/cmr/timeseries.py
hrodmn: Yes that is a great point. I was thinking about adding some kind of explainer in the notebook but you are right - at no point do we perform an average over multiple assets in a time range. The [`mosaic_reader`](https://github.com/cogeotiff/rio-tiler/blob/9a39c802044b485a26c29065b539ec44c817d8a6/rio_tiler/mosaic/reader.py#L25) can take a list of assets but they get flattened to a single 2D array (per variable) and I am 90% sure it will just fill in the array one asset at a time, and if they overlap the final values come from the last asset.

I will add some more annotations to the arguments here and some details in the notebook to explain that better.

titiler/cmr/timeseries.py
```diff
@@ -0,0 +1,612 @@
...
+        Query(
+            description="End datetime for timeseries request",
+        ),
+    ] = None
+    step: Annotated[
+        Optional[str],
+        Query(
+            description="Time step between timeseries intervals, expressed as [ISO 8601 duration](https://en.wikipedia.org/wiki/ISO_8601#Durations)"
+        ),
+    ] = None
```
maxrjones: What do you think of consolidating the time related parameters to match the syntax in the tiles enpoint in titiler-cmr?

image

https://github.com/developmentseed/titiler-cmr/blob/afe8b6dee5d2202fd8213080f986f3b66125d8b6/titiler/cmr/dependencies.py#L98-L112

maxrjones (commented):
I'm totally blown away by the GIF possibilities 🎉 🌍 I unfortunately wasn't able to fully dive into a review today, so I will try it out more on Monday but wanted to share a couple questions/comments so far. 

My main question is about the pros/cons of having a new `/timeseries/bbox` url rather than adding support for returning `image/gif` to the `images/bbox` url. They share a ton of parameters and to me seem conceptually very similar with the return type just dependent on whether you have discrete mosaicing in time. If it remains under `/timeseries` and will be specific to gif return types, would it help for clarity to just name if `timeseries/gif` rather than `timeseries/bbox`?

titiler/cmr/timeseries.py
hrodmn: I like that idea! We could at least collapse the start_datetime and end_datetime parameters into one `datetime` parameter and split on `/`. I'll have to think a bit more about how we would encode the other variables into a single parameter but maybe there is a convention out there somewhere.

hrodmn:
> My main question is about the pros/cons of having a new `/timeseries/bbox` url rather than adding support for returning `image/gif` to the `images/bbox` url. They share a ton of parameters and to me seem conceptually very similar with the return type just dependent on whether you have discrete mosaicing in time. 

I think we could just extend the original `/bbox` endpoint but there would wind up being a big `if "gif" then ... else: ...` control flow since the timeseries functionality is just farming out the requests to the `/bbox` endpoint again. Maybe there would be an elegant way to write a recursive approach.

> If it remains under `/timeseries` and will be specific to gif return types, would it help for clarity to just name if `timeseries/gif` rather than `timeseries/bbox`?

This would definitely make sense if we are just going to stick with the `gif` format but I want to add support for `mp4` export, too, so I kept the `format` argument in there.

I think the `/bbox` endpoint label does a bad job describing what it actually does, so maybe it would make more sense to do `/timeseries/{format}` where `format` could be `gif` or `mp4`.

titiler/cmr/timeseries.py
vincentsarago: >  and if they overlap the final values come from the last asset.

From the first asset! (the default mosaic selection method is `FirstMethod` https://github.com/cogeotiff/rio-tiler/blob/9a39c802044b485a26c29065b539ec44c817d8a6/rio_tiler/mosaic/methods/defaults.py#L10-L27

Dockerfile
```diff
@@ -26,5 +26,5 @@ RUN if [ -z "$EARTHDATA_USERNAME" ] || [ -z "$EARTHDATA_PASSWORD" ]; then \
 # http://www.uvicorn.org/settings/
 ENV HOST 0.0.0.0
 ENV PORT 80
-CMD uv run uvicorn titiler.cmr.main:app --host ${HOST} --port ${PORT} --log-level debug
+CMD uv run uvicorn titiler.cmr.main:app --host ${HOST} --port ${PORT} --log-level debug --reload
```
vincentsarago, line 29: you don't really need the `reload` here because the code is `copied` not mounted

pyproject.toml
```diff
@@ -37,6 +37,9 @@ dependencies = [
     "s3fs~=2024.9.0",
     "xarray~=2024.9.0",
     "geojson-pydantic>=1.1.1",
+    "python-dateutil>=2.9.0.post0",
+    "httpx>=0.27.2",
+    "pillow>=11.0.0",
```
vincentsarago, line 41: I've never tried but it seems that GDAL can create GIF natively https://gdal.org/en/latest/drivers/raster/gif.html 

I fear that adding pillow dependency will make the package size 💥 (we could also remove some dependencies like S3FS)

maybe we should ask @kylebarron to add python bindings to https://github.com/image-rs/image-gif 🙈

pyproject.toml
```diff
@@ -45,10 +48,11 @@ uvicorn = [
     "uvicorn",
 ]
 dev = [
-    "pre-commit",
-    "jupyterlab>=4.2.5",
-    "httpx",
     "folium",
+    "httpx",
```
vincentsarago: we can remove httpx from the `dev` and `test` dependencies list

pyproject.toml
```diff
@@ -57,6 +61,7 @@ test = [
     "httpx>=0.27.2",
     "pytest-mock>=3.14.0",
     "pytest-recording>=0.13.2",
+    "asgi-lifespan>=2.1.0",
```
vincentsarago: is that used somewhere?

tests/test_app.py
```diff
@@ -220,13 +228,179 @@ def test_xarray_part(
...
+            },
+        )
+    assert response.status_code == 200
+    assert response.headers["content-type"] == "image/png"
+
+    image_data = io.BytesIO(response.content)
+    Image.open(image_data)
+
+    # Check dimensions
+    # assert image.size == size, f"Expected image size {size}, but got {image.size}"
```
vincentsarago, line 256: 👀

tests/test_app.py
```diff
@@ -220,13 +228,179 @@ def test_xarray_part(
...
+            ],
+            "valid_percent": 93.72,
+            "masked_pixels": 301.0,
+            "valid_pixels": 4493.0,
+            "percentile_2": 0.0,
+            "percentile_98": 0.99,
+        }
+    }
+
+    async def mock_timestep_request(url: str, **kwargs) -> Response:
```
vincentsarago, line 297: Note: maybe over complex but there might be way to create a proper app running in threads using https://github.com/banesullivan/server-thread (https://github.com/developmentseed/rio-viz/blob/776fb29e5ce2647cdb39d962e9bb0de833f74feb/rio_viz/app.py#L799-L818)

tests/test_timeseries.py
```diff
@@ -0,0 +1,274 @@
...
+    assert parse_duration("P1Y") == relativedelta(years=1)
+    assert parse_duration("P2M") == relativedelta(months=2)
+    assert parse_duration("P3D") == relativedelta(days=3)
+    assert parse_duration("PT4H") == relativedelta(hours=4)
+    assert parse_duration("PT5M") == relativedelta(minutes=5)
+    assert parse_duration("PT6S") == relativedelta(seconds=6)
+    assert parse_duration("PT1S") == relativedelta(seconds=1)
+    assert parse_duration("P1Y2M3DT4H5M6S") == relativedelta(
+        years=1, months=2, days=3, hours=4, minutes=5, seconds=6
+    )
```
vincentsarago: you can use `pytest.mark.parametrize` 

https://github.com/cogeotiff/rio-tiler/blob/9a39c802044b485a26c29065b539ec44c817d8a6/tests/test_expression.py#L15-L27

tests/test_timeseries_extension.py
```diff
@@ -0,0 +1,23 @@
...
+    tiler_plus_timeseries = Endpoints(extensions=[TimeseriesExtension()])
+    # Check that we added two routes
+    assert len(tiler_plus_timeseries.router.routes) > len(tiler.router.routes)
+
+    timeseries_app = FastAPI()
+    timeseries_app.include_router(tiler_plus_timeseries.router)
+
+    assert any(
+        "/timeseries/statistics" in str(route) for route in timeseries_app.router.routes
+    )
```
vincentsarago, line 21: I don't think we need to run this async

titiler/cmr/timeseries.py
```diff
@@ -0,0 +1,612 @@
...
+
+class TimeseriesImageType(str, Enum):
+    """Available output image type."""
+
+    gif = "gif"
+
+    @DynamicClassAttribute
+    def mediatype(self):
+        """Return image media type."""
+        return TimeseriesMediaType[self._name_].value
```
vincentsarago, line 78: TODO: remove once we update titiler >=0.19 https://github.com/developmentseed/titiler/pull/1018

tests/test_app.py
hrodmn: I added this test then realized that the `XarrayReader` in rio-tiler did not support height/width arguments yet, but I think you added those in the last release so I'll fix this test when we handle the upgrade to rio-tiler and titiler.

pyproject.toml
kylebarron: > we could also remove some dependencies like S3FS

Yes!
 
> maybe we should ask @kylebarron to add python bindings to [image-rs/image-gif](https://github.com/image-rs/image-gif) 🙈

More rust!!

maxrjones:
> > My main question is about the pros/cons of having a new `/timeseries/bbox` url rather than adding support for returning `image/gif` to the `images/bbox` url. They share a ton of parameters and to me seem conceptually very similar with the return type just dependent on whether you have discrete mosaicing in time.
> 
> I think we could just extend the original `/bbox` endpoint but there would wind up being a big `if "gif" then ... else: ...` control flow since the timeseries functionality is just farming out the requests to the `/bbox` endpoint again. Maybe there would be an elegant way to write a recursive approach.
> 
> > If it remains under `/timeseries` and will be specific to gif return types, would it help for clarity to just name if `timeseries/gif` rather than `timeseries/bbox`?
> 
> This would definitely make sense if we are just going to stick with the `gif` format but I want to add support for `mp4` export, too, so I kept the `format` argument in there.
> 
> I think the `/bbox` endpoint label does a bad job describing what it actually does, so maybe it would make more sense to do `/timeseries/{format}` where `format` could be `gif` or `mp4`.

The `timeseries/bbox` url calls the `images/bbox` url to get the individual images to combine into a gif, right? If so, it might be safer to keep them separate to avoid a potential for recursive lambda calls. I would lean towards a more descriptive label like `/timeseries/{format}` or `/timeseries/movie` with two output types.

pyproject.toml
hrodmn: It was at some point but I wound up not needing it.

titiler/cmr/timeseries.py
hrodmn: I found a library (`isodate`) that handles parsing durations so this function is gone now.

titiler/cmr/timeseries.py
hrodmn: I wound up consolidating it down to `datetime`, `step` and `temporal_mode` (formerly (`exact`). `datetime` can now accommodate a comma-separated list of datetimes/intervals (or just a single datetime/interval without any commas).

pyproject.toml
hrodmn: I think GDAL can only write a single image to a GIF. I tried making a vrt out of two PNGs and then ran `gdal_translate` with `-of GIF` and got an error:
```bash
$ gdalbuildvrt -separate /tmp/test.vrt /tmp/2023-11-01T00\:00\:01Z.png /tmp/2024-05-15T00\:00\:01Z.png
0...10...20...30...40...50...60...70...80...90...100 - done.

$ gdal_translate -of GIF /tmp/test.vrt /tmp/test.gif
Input file size is 1440, 720
ERROR 6: GIF driver only supports one band images.
```

I ran `cdk synth` locally and compared the size of the zipped asset outputs in `cdk.out/` on the `develop` branch (without the `Pillow` dependency) and on this branch (with `Pillow`) and it looks like it only increases the size of the zipped code package by ~4MB so it might be ok.

@vincentsarago would it be ok if we keep `Pillow` in this PR and see how it looks when we deploy it? It would be great to keep it slim with `image-rs/image-gif` but if `Pillow` works for now that would be ideal.

vincentsarago: >  it looks like it only increases the size of the zipped code package by ~4MB so it might be ok.

🙏 looks ok then :-)

tests/test_timeseries.py
hrodmn: I wound up finding a lightweight import (`isodate`) that has a `parse_duration` function that we can use instead of a home cooked regex function.

tests/test_app.py
hrodmn: In this case I am comfortable with relying on the mocked PNG responses since we have already have tests for the `/bbox` endpoint.

Dockerfile
hrodmn: I added it because I mount the `titiler/` directory in docker-compose.yml. It is redundant to `COPY` and mount the volume but it is convenient for local development to have the reload functionality.

titiler/cmr/timeseries.py
hrodmn: @abarciauskas-bgse yes that is correct! The code in the `/timeseries` endpoints just converts the timeseries query parameters into a list of `datetime` parameters, then a fires off a lower-level request for each unique `datetime` parameter. Then there is a little bit of code to organize the results (e.g. list of PNGs to a GIF, list of statistics geojson output into a single geojson).

hrodmn:
> I would lean towards a more descriptive label like /timeseries/{format} or /timeseries/movie with two output types.

I went with `/timeseries/bbox/{xmin},{ymin},{xmax},{ymax}.{format}` to be consistent with the non-timeseries `/bbox` endpoint, but I agree with @maxrjones that `/timeseries/bbox` path makes it less than obvious what it actually does. @vincentsarago do you have an opinion on staying consistent with `/bbox` or more creating a more descriptive with `/timeseries/{format}` (-> `/timeseries/gif`, `/timeseries/mp4`)?

maxrjones:
> > I would lean towards a more descriptive label like /timeseries/{format} or /timeseries/movie with two output types.
> 
> I went with `/timeseries/bbox/{xmin},{ymin},{xmax},{ymax}.{format}` to be consistent with the non-timeseries `/bbox` endpoint, but I agree with @maxrjones that `/timeseries/bbox` path makes it less than obvious what it actually does. @vincentsarago do you have an opinion on staying consistent with `/bbox` or more creating a more descriptive with `/timeseries/{format}` (-> `/timeseries/gif`, `/timeseries/mp4`)?

sounds good, I think this could be considered a documentation issue. the self-describing APIs are nice for maintenance, but a downside is new people may not find it sufficient for finding features.

titiler/cmr/timeseries.py
hrodmn: I actually wound up dropping the `step_idx` parameter in pursuit of a more succinct API.

hrodmn:
Thank you all for your thoughtful reviews of these changes!!

After thinking it over some more I want to stick with `/timeseries/bbox` for the sake of consistency with other titiler APIs. I added some more verbose descriptions to the openapi documentation to make the purpose of each `/timeseries` endpoint more clear. I also opened up an issue to actually publish the docs to the web! #38

PR #42, opened 2024-12-03
Deploy documentation and document time series api limits

hrodmn:
This PR adds changes for a) deploying the documentation to GH pages, b) adds an API benchmarking process (and report in the docs) for the time series endpoints. After talking about it some more with the team I think we can be more surgical in the benchmark approach (i.e. predict limits based on Lambda configuration parameters), but the exercise was still helpful for understanding when things fall apart and which knobs you can turn to get time series requests that return a response before the Lambda breaks down. There are definitely some modifications or alternative approaches that are worth considering to improve the capabilities of this application!

A preview of the docs for this PR are available here: https://developmentseed.org/titiler-cmr/pr-previews/pr-42/

I'm not sure if I'll keep the PR preview feature alive since it is so easy to deploy the docs locally (`uv sync && uv run mkdocs serve -o`) but I'll leave it up for now so others can take a look at the rendered site.


docs/examples/time_series_example.ipynb
```diff
@@ -0,0 +1,514 @@
+{
```
abarciauskas-bgse, line 1: <p>What do you mean by "time-aware responses" (I think this reads well without that first sentence)</p><p> </p><p>Also not sure what is meant by "time series structure" - does this just mean those are the required parameters to construct a time series request?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/developmentseed/titiler-cmr/pull/42/discussion/#comment-1871974777'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='0'/>

abarciauskas-bgse, line 1: <p>For the step parameter can we add an example? And perhaps link to <a href="https://www.digi.com/resources/documentation/digidocs/90001488-13/reference/r_iso_8601_duration_format.htm" target="_blank">https://www.digi.com/resources/documentation/digidocs/90001488-13/reference/r_iso_8601_duration_format.htm</a> ?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/developmentseed/titiler-cmr/pull/42/discussion/#comment-1871974779'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='3'/>

docs/time_series_performance_benchmarks.ipynb
```diff
@@ -0,0 +1,156 @@
+{
```
abarciauskas-bgse: <p>Should we mention the dataset used here? Does that impact the number of time points which can be returned in under 30 seconds?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/developmentseed/titiler-cmr/pull/42/discussion/#comment-1871981222'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='0'/>

abarciauskas-bgse: <p>Perhaps also mention 28km ~ 0.25 degrees?</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/developmentseed/titiler-cmr/pull/42/discussion/#comment-1871981226'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='2'/>

abarciauskas-bgse: <p>I think this sentence isn't complete, should it say "so reducing the size of the AOI makes little impact on run time"</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/developmentseed/titiler-cmr/pull/42/discussion/#comment-1871981227'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='5'/>

abarciauskas-bgse: <p>Are the bbox dimensions the size in pixels or degrees? I'm assuming pixels but then curious why we have 360x180 when the other options are square and also a factor of standard tile sizes, in pixels.</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/developmentseed/titiler-cmr/pull/42/discussion/#comment-1871981228'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='7'/>

abarciauskas-bgse: <p>Suggested rewording: "The response time increases with the size of the area of interest, especially for higher resolution images"</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/developmentseed/titiler-cmr/pull/42/discussion/#comment-1871981230'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='9'/>

abarciauskas-bgse:
Thanks @hrodmn ! I'm still trying to parse everything that is here but most importantly I want to understand how to interpret the time series benchmarks:

1. It seems like for statistics endpoints (e.g. creating a time series as a json response) the AOI doesn't matter since the whole granule has to be read.
    a. I'm wondering if this is always the case, like if the granule is chunked in space?
    b. the limit of 1000 points is for this specific datasets' resolution (~0.25 degrees) right?
2. For the bbox gif endpoint, it sounds like the size of the AOI does matter but is it more generally the size of the returned image, e.g. 512x512 vs 1024x1024?
    a. Similar question for the limit of 500, this is for the specific dataset used?

hrodmn:
After talking about this with the team a few weeks ago I think it does make sense to be more surgical in the benchmarking tests rather than throwing the massive requests at it when we know there will be an error. The Lambda configuration has a concurrency limit of 998, and since the endpoints start to get unstable beyond 500 time points (due to other resource constraints) I think it makes sense to set a maximum time series length of 500 while we consider how/if we want to make the system more efficient.

> 1. It seems like for statistics endpoints (e.g. creating a time series as a json response) the AOI doesn't matter since the whole granule has to be read.
>    a. I'm wondering if this is always the case, like if the granule is chunked in space?

I think you are correct that it depends on the chunking scheme of the dataset. All of this happens at the rio-tiler level, here is the moment where a full dataset gets cropped down to the bounding box in the `part` method:
https://github.com/cogeotiff/rio-tiler/blob/83e8fc4ed765445d1aa0267cc3bccee15664b9b3/rio_tiler/io/xarray.py#L301-L305

I will do some testing on this but I hope that if there were some spatial chunking the full dataset is not loaded into memory in order to do that crop operation.

>    b. the limit of 1000 points is for this specific datasets' resolution (~0.25 degrees) right?

I am adding some tests that use the MUR-SST dataset and I still need to drill into the details but the limits are much lower for that dataset so I will try to update guidance for users based on the resolution of the input dataset.

> 2. For the bbox gif endpoint, it sounds like the size of the AOI does matter but is it more generally the size of the returned image, e.g. 512x512 vs 1024x1024?
>    a. Similar question for the limit of 500, this is for the specific dataset used?

Yes the resolution of the output image has an impact on the reduce step in the map/reduce process. If we are trying to combine many high resolution images in the original time series request I think we are likely to hit memory and/or timeout limits. I need to go to the Lambda logs to figure out exactly what is going on, though.

abarciauskas-bgse:
@hrodmn thanks for this, this guidance is super helpful. 2 questions:

1. Do you mind if I change `pixels` to `grid_cells` in all places where we're referring to the original data? When I read pixels I think of the resolution of the output image but I think we are mostly referring to the input data in this guidance.
2. Do you have any sense why the individual image size matters aside from it's contribution to the total image size?

hrodmn:
> @hrodmn thanks for this, this guidance is super helpful. 2 questions:
> 
> 1. Do you mind if I change `pixels` to `grid_cells` in all places where we're referring to the original data? When I read pixels I think of the resolution of the output image but I think we are mostly referring to the input data in this guidance.

Sure that makes sense to me - is that a convention for netcdf datasets? Most of my experience is with raster datasets so I usually think about cells as pixels.

> 2. Do you have any sense why the individual image size matters aside from it's contribution to the total image size?

I have only done a small amount of profiling in the Lambda context, but I think there are two ways large images can affect time series requests:
- Each lower-level request requires more resources (time + memory) which increases the probability that not all requests will succeed within 30 seconds or overflow the Lambda memory limit. I did some rough testing to figure out the maximum size for a single image that the Lambda could process without crashing and used that to set the cap for individual image sizes.
- Some kind of resource limitation within the initial Lambda function call (e.g. /timeseries/bbox) makes the request for large images take longer (this is only a half-formed thought). I have tried using my local environment as a 'scheduler' by running the docker network locally with lower-level requests redirected to the actual deployed API (instead of the local API). When I do this I can make larger requests work, so I think there is something specific to the Lambda deployment that is limiting the capacity.

abarciauskas-bgse:
@hrodmn 
re `grid cells` vs `pixels` - I have come across pixels being used in place of cells for raster datasets and always found it confusing. I think grid cell is common term for netcdf data but can't say declaratively. This all may just be a matter of personal preference so if you prefer pixels, perhaps just including a line about terminology may be helpful. Looking at some of the docs for NetCDF I just see references to values and cells...

And thanks for the responses to question 2 - those reasons make sense to me. I don't think it's worth further investigation at this point and it's great that the individual image size limits are included in the documentation.

PR #48, opened 2025-05-21
Fix/fixes for veda deploy

abarciauskas-bgse:
For some reason, boto3 was not installed without the `run` in this command. I can't explain it but it appears to work with this part of the command included.

Successful job using this branch: https://github.com/NASA-IMPACT/veda-deploy/actions/runs/15173444981/job/42668957562, failed job using develop branch: https://github.com/NASA-IMPACT/veda-deploy/actions/runs/15173235029/job/42668123921

### DIFF
--- a/.github/actions/cdk-deploy/action.yml
+++ b/.github/actions/cdk-deploy/action.yml
@@ -51,14 +51,13 @@ runs:
       run: |
         uv sync --only-group deployment
         uv run npm install
-        uv run pip install boto3 # for the generate_env_file.py script
 
     - name: Get relevant environment configuration from aws secrets
       if: inputs.env_aws_secret_name != ''
       shell: bash
       env:
         AWS_DEFAULT_REGION: us-west-2
-      run: python ${{ inputs.script_path }} --secret-id ${{ inputs.env_aws_secret_name }} --env-file ${{ inputs.dir }}/.env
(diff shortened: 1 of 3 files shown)


hrodmn (commented):
This seems to work so if you want to leave it here that's fine but I think the right thing to do is to add `boto3` to the deployment dependency group (`uv add --group deployment boto3`) then update the env config step like this:
```python
    - name: Get relevant environment configuration from aws secrets
      if: inputs.env_aws_secret_name != ''
      shell: bash
      env:
        AWS_DEFAULT_REGION: us-west-2
      run: uv run ${{ inputs.script_path }} --secret-id ${{ inputs.env_aws_secret_name }} --env-file ${{ inputs.dir }}/.env
```

The key change is changing `python ${{ inputs.script_path }}` to `uv run ${{ inputs.script_path }}`. 

Since the env config step is not using `uv run` to execute the script from `input.script_path`, it is using some other python environment! I guess I got lucky that it worked for titiler-multidim (still don't know exactly why).

abarciauskas-bgse:
That makes a lot of sense, thank you @hrodmn 

~I think I will actually try to add boto3 to the deployment group in pyproject.toml~ I realize this is what you meant by `uv add --group deployment boto3`

abarciauskas-bgse:
🤔 this still isn't working (see run https://github.com/NASA-IMPACT/veda-deploy/actions/runs/15191552245/job/42726049833)

PR #50, opened 2025-05-23
Fix/separate app stack settings

abarciauskas-bgse:
Add root_path and veda_custom_host to replicate https://github.com/developmentseed/titiler-multidim/pull/87/

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -17,7 +17,7 @@ services:
       - HOST=0.0.0.0
       - PORT=8081
       - WEB_CONCURRENCY=1
-      - TITILER_CMR_API_DEBUG=TRUE
+      - TITILER_CMR_DEBUG=TRUE
       - TITILER_CMR_S3_AUTH_STRATEGY=iam
       - TITILER_CMR_S3_AUTH_ACCESS=external
       # earthdata authentication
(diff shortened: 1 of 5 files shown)


abarciauskas-bgse:
I have verified this is now working via [this tile request](https://staging.openveda.cloud/api/titiler-cmr/WebMercatorQuad/map?concept_id=C3273639213-GES_DISC&datetime=2018-02-12T09%3A00%3A00Z&variable=NPP&backend=xarray&rescale=0%2C7.191663087269262e-080) and visiting the API docs landing page.

It does seem a bit duplicative that TITILER_CMR_ROOT_PATH is defined by AppSettings in infrastructure/aws/config.py in order to be passed as an environment variable to the lambda run time for use ApiSettings in titiler/cmr/main.py but I haven't thought of a better way to pass environment variables between the stack deployment and the app run time...

hrodmn (approved):
> It does seem a bit duplicative that TITILER_CMR_ROOT_PATH is defined by AppSettings in infrastructure/aws/config.py in order to be passed as an environment variable to the lambda run time for use ApiSettings in titiler/cmr/main.py but I haven't thought of a better way to pass environment variables between the stack deployment and the app run time...

Agreed. The number of variable handoffs is a little out of control especially when you go all the way up to the veda-deploy level! I did not try experimenting with the whole custom host and root path situation when I did this in titiler-multidim - I just followed some example from a different veda service. 

I guess we could eliminate the infrastructure AppSettings and just load things directly from env vars at deploy time, but the pydantic settings are a convenient interface for validation and defining defaults. I think we are stuck with the env var -> settings in the titiler-cmr application since we don't compile the app to encode these variables in the runtime.

Issue #49, opened 2025-05-21
Outstanding TODOs for veda-deploy


abarciauskas-bgse:
veda-deploy's smce-staging environment uses the `STAGING_Deployment_Envs` secret, which includes a `STAGE` variable (set to "staging"). We don't want to use the value `staging` for the titiler-cmr CDK stack's stage yet, as the current "titiler-cmr-staging" cloudformation deployment is in use as `dev-titiler-cmr.delta-backend.com` (for [hls](https://github.com/NASA-IMPACT/veda-config/blob/develop/datasets/hls_2.0.data.mdx) and [gpm imerg](https://github.com/NASA-IMPACT/veda-config/blob/develop/datasets/GPM_3IMERGDF.data.mdx).

Note `StackSettings` in config.py expects the `TITILER_CMR_` prefix and currently the `TITILER_CMR_STAGE` is set to `test` in `STAGING_Deployment_Envs`. If this is removed it will default to `production`, which is not a big deal because I don't believe the titiler-cmr-production stack is in use and is quite behind. To smoothly transition to using the `STAGE` environment variable the following is proposed:

- [x] Continue to use the `TITILER_CMR_STAGE` environment variable set to `test`. Once the `test` deployment is working from veda-deploy (✅ ), create a https://staging.openveda.cloud/api/titiler-cmr/ endpoint and point it at the new veda-deploy managed deployment.
- [x] Update the configuration in veda-config to point to https://staging.openveda.cloud/api/titiler-cmr/.
... Wait for release ...
- [x] test once new release is live
- [x] Update documentation in titiler-cmr
- [x] Delete the existing titiler-cmr-staging stack.
- [x] Make the `stage` setting in infrastructure/aws/cdk/config.py set via environment variable `STAGE` instead of `TITILER_CMR_STAGE` to be consistent with other veda-deploy deployments.
- [x] Tag a new version which includes the deployment code and update TITILER_CMR_GIT_REF in veda-deploy
- [x] Redeploy titiler-cmr deployment in veda-deploy. This will be the new `titiler-cmr-staging` stack. Once tested, point https://staging.openveda.cloud/api/titiler-cmr/ at this stack and delete `titiler-cmr-test`.
- [x] (Unrelated) to the above: Delete the existing prod stack

PR #51, opened 2025-06-06
Feat/add dev deploy

abarciauskas-bgse:
Adds a deployment to a development stack for testing features. Note I still plan to check off all clean up items in https://github.com/developmentseed/titiler-cmr/issues/49, but felt having a stack just for testing new features would be useful.


hrodmn (commented):
Great idea to make it possible to safely deploy from a branch! My only question: what is the use-case for running this in CI on a pull request instead of just manually triggering it?

abarciauskas-bgse:
@hrodmn In the case of testing a branch which adds new features, I thought it would be convenient to have the ability to label the associated PR and then upon any push to that branch (PR synchronize) the deploy dev action would be triggered. But if you think it's more prudent to just have the dispatch trigger, I'm ok with that too.

abarciauskas-bgse:
@hrodmn just checking - do you think this PR is good to merge?

hrodmn (approved):
Yes let's merge it!

PR #52, opened 2025-06-09
feat: incorporate xarray dimension selection

abarciauskas-bgse:
This WIP PR addresses https://github.com/NASA-IMPACT/veda-odd/issues/188 so that we can select various additional dimensions' values (specifically `time` and `lev`) when a netcdf has more than 2 dimensions. Each file for the TROPESS O3 dataset has 4 dimensions.

For this we can use @vincentsarago's extremely helpful work to add the `sel` and `sel_method` parameters to the titiler.xarray reader. I incorporated those parameters using the dependencies module in titiler.xarray and removed `ZarrParams`.

I also moved tests around a bit to make them easier to navigate.


titiler/cmr/factory.py
```diff
@@ -50,6 +49,7 @@
 from titiler.core.resources.enums import ImageType, OptionalHeader
 from titiler.core.resources.responses import GeoJSONResponse
 from titiler.xarray.io import Reader as XarrayReader
+from titiler.xarray.dependencies import XarrayIOParams, CompatXarrayParams
```
hrodmn, line 52: :100: - it's so nice to be able to import more from titiler.xarray for this type of thing

pyproject.toml
```diff
@@ -22,9 +22,9 @@ classifiers = [
     "Topic :: Scientific/Engineering :: GIS",
 ]
 dependencies = [
-    "titiler.core>=0.21.0,<0.22",
-    "titiler.mosaic>=0.21.0,<0.22",
-    "titiler.xarray>=0.21.0,<0.22",
+    "titiler.core>=0.22.0,<0.23",
+    "titiler.mosaic>=0.22.0,<0.23",
+    "titiler.xarray@git+https://github.com/developmentseed/titiler.git@ab/add-sel-by-datetime-to-xarray#subdirectory=src/titiler/xarray",
```
vincentsarago:
```suggestion
    "titiler.core>=0.22.3,<0.23",
    "titiler.mosaic>=0.22.3,<0.23",
    "titiler.xarray>=0.22.3,<0.23",
```

tests/test_xarray_backend.py
```diff
@@ -0,0 +1,405 @@
...
+
+    assert response.status_code == 200
+    assert response.headers["content-type"] == "application/json"
+
+    tilejson = response.json()
+    assert tilejson["bounds"] == [-180.0, -90.0, 180.0, 90.0]
+
+
+@pytest.mark.vcr
+def test_xarray_tilejson_with_sel(app, xarray_query_params):
```
abarciauskas-bgse, line 41: I added this test, otherwise tests are unchanged apart from moving them to separate files (test_rasterio_backend and test_xarray_backend)

titiler/cmr/reader.py
```diff
@@ -123,68 +123,6 @@ def xarray_open_dataset(
     return ds
 
 
-def arrange_coordinates(da: xarray.DataArray) -> xarray.DataArray:
```
abarciauskas-bgse, line 126: These functions are unused, as far as I can tell

hrodmn (approved):
This looks great! The only feature I think we should consider adding is some mechanism for supporting `sel` in the time series endpoints, but I don't think we need to do that in this PR.

Right now titiler-cmr will convert a time series request into a series of requests with different values for the `datetime` parameter, but there are cases (TROPESS) where we would also need to use the `sel` parameter to choose a slice from the time dimension of a multi-temporal granule. Maybe a flag like `sel_time_with_datetime` could signal titiler-cmr to also pass a `sel=time={datetime}` parameter.

PR #58, opened 2025-06-24
Fix/environment vars

abarciauskas-bgse:
This PR updates some environment variables required by titiler-cmr and provides documentation following the guidance in https://github.com/NASA-IMPACT/veda-deploy/pull/125#issuecomment-2993074242

Successful run: https://github.com/NASA-IMPACT/veda-deploy/actions/runs/15791308215/job/44517296320 using this branch created an updated titiler-cmr-staging deployment with the AWS_REQUEST_PAYS=requester and TITILER_CMR_S3_AUTH_STRATEGY variables set in the lambda environment.


infrastructure/aws/cdk/app.py
```diff
@@ -76,8 +74,9 @@ def __init__(
             timeout=Duration.seconds(timeout),
             environment={
                 **DEFAULT_ENV,
-                **environment,
                 "TITILER_CMR_ROOT_PATH": app_settings.root_path,
+                "TITILER_CMR_S3_AUTH_STRATEGY": app_settings.s3_auth_strategy,
+                "AWS_REQUEST_PAYER": app_settings.aws_request_payer,
```
hrodmn: Should we only add this `AWS_REQUEST_PAYER` if `app_settings.aws_request_payer is not None`?

hrodmn (approved):
I was confused at first about why the env vars got added to deploy.yml but then I remembered that veda-deploy does not use that file and uses cdk-deploy/action.yml instead. Thanks for making this change!

PR #62, opened 2025-06-25
fix: handle case where UMM metadata is missing horizontal resolution

### LINKED ISSUE #59: How to solve for error when `ResolutionAndCoordinateSystem` does not exist in umm json
It looks like the 2 examples used in the documentation both have `ResolutionAndCoordinateSystem` present in UMM metadata, but when not present, as is the case in many datasets, timeseries generation will fail because the application will get a key error in [`get_resolution_degrees`](https://github.com/developmentseed/titiler-cmr/blob/develop/titiler/cmr/utils.py#L80-L86)

Perhaps if this `ResolutionAndCoordinateSystem` key is not present we have no way of doing the nice request size estimation and should allow the app to just try to generate the time series? I'm not sure of another way to get the dataset resolution without opening one of the files.

What do you think @hrodmn ?

hrodmn:
There are concept_ids which are missing the `HorizontalDataResolution` attribute so we cannot calculate the array size from collection-level metadata alone. This change will allow any time series request for a dataset that does not include this piece of information to proceed regardless of request size.

resolves #59

### DIFF
--- a/titiler/cmr/utils.py
+++ b/titiler/cmr/utils.py
@@ -80,7 +80,7 @@ def parse_datetime(
     return datetime_, start, end
 
 
-def get_resolution_degrees(concept_id: str) -> Tuple[float, float]:
+def get_resolution_degrees(concept_id: str) -> Tuple[Optional[float], Optional[float]]:
     """Query CMR to get the resolution of a dataset using its concept_id. If the units are in meters
     convert to degrees using the rough conversion factor of 0.00001 degrees per meter"""
     ds = earthaccess.collection_query().concept_id(concept_id).get()[0]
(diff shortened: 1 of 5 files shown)


titiler/cmr/utils.py
```diff
@@ -81,9 +84,17 @@ def get_resolution_degrees(concept_id: str) -> Tuple[float, float]:
...
+
+    try:
+        resolution_info = ds["umm"]["SpatialExtent"]["HorizontalSpatialDomain"][
+            "ResolutionAndCoordinateSystem"
+        ]["HorizontalDataResolution"]["GenericResolutions"][0]
+    except KeyError:
+        logger.warning(
+            f"could not find HorizontalDataResolution for concept_id {concept_id}"
+        )
+        return (0, 0)
```
abarciauskas-bgse: Do you prefer returning `0` to `None` here? The latter feels more appropriate given it's not that the resolution is ever `0` but unknown (unfortunately unknown is not a type 😅 )

hrodmn: `None` would make more sense, I think I was just being lazy about typing. I'll change it!

PR #63, opened 2025-06-25
Add requester pays for s3fs

abarciauskas-bgse:


hrodmn (approved):
Thanks for sorting this out! It is too bad that the env var is not used by s3fs. Maybe we can file an issue to update this module to use obstore instead?

PR #68, opened 2025-07-02
Fix/upgrade geojson pydantic

abarciauskas-bgse:
I verified that tests do in fact fail on the develop branch when geojson_pydantic is version 2.0.0. With the change in this PR (just call `.iter()` in geojson_statistics), we can use the upgraded geojson_pydantic.

For good measure I added a test for `FeatureCollection` objects passed to /statistics since only a single `Feature` was covered, however this addition is not strictly necessary for this change.

The only other place I can tell that a FeatureCollection is iterated upon is in [get_geojson_bounds](https://github.com/developmentseed/titiler-cmr/blob/develop/titiler/cmr/utils.py#L157), which already uses `fc.features` which I believe is still supported in the upgraded goejson_pydantic.


pyproject.toml
```diff
@@ -36,10 +36,10 @@ dependencies = [
     "rioxarray~=0.13.4",
     "s3fs~=2024.9.0",
     "xarray~=2024.9.0",
-    "geojson-pydantic>=1.1.1",
+    "geojson-pydantic>=2.0.0,<3.0",
     "python-dateutil>=2.9.0.post0",
     "httpx>=0.27.2",
-    "pillow>=11.0.0",
+    "pillow==11.2.1",
```
abarciauskas-bgse: I was getting a pillow build error, which may be related to the release yesterday? Looking at this issue: https://github.com/python-pillow/Pillow/issues/9062, the [titiler-cmr build failures are similar](https://github.com/developmentseed/titiler-cmr/actions/runs/16030522087/job/45229930367)

abarciauskas-bgse: it looks like this issue was resolved by `uploaded manylinux2014 wheels to PyPI`

hrodmn (approved):
This looks great - thanks @abarciauskas-bgse!

PR #76, opened 2025-09-22
feat: add parameters to pass datetime through to sel in time series

### LINKED ISSUE #77: update time series endpoints to accommodate sel=time case
Users can reach into particular slices of granules by using the `sel` parameter for the xarray backend. To enable this type of access in the time series endpoints we need to add a way for users to signal titiler-cmr to include the datetime parameter in the form of `sel=time={datetime}`.

hrodmn:
This adds a pair of new parameters to the /timeseries endpoints: `use_sel_for_datetime` and `sel_time_dim`. These parameters allow users to have the `datetime` values get passed to the sel parameter in the lower-level requests like this: `sel={sel_time_dim}={datetime}`. This unblocks the use-case where granules cover a datetime range and have a time dimension within each file.

I also had to update the test cassettes for reasons that I don't fully understand, but I successfully created a fixture that actually routes the lower-level time series requests to the test app instead of just mocking that response (which was important to test this new complex case).


hrodmn:
Closing in favor of #78

PR #79, opened 2025-09-29
Add obstore-backed Xarray reader for TiTiler-CMR

negin513:
### Summary

This PR adds an obstore-based data access path for TiTiler-CMR/xarray workloads. It enables opening NetCDF (HDF5) and Zarr datasets over S3, HTTPS, and local file URLs without fsspec/s3fs, and wires in Earthdata S3 temporary credentials via earthaccess.

It removes fsspec/s3fs as hard dependencies and adds obstore as a dependency.

Note: VirtualiZarr is not included here. Also, we’re avoiding extra dependencies for now (particularly due to deployment limits in AWS Lambda).


hrodmn:
This feature landed in #145 so I'm closing this PR

PR #81, opened 2025-10-15
feat: add OpenTelemetry + X-Ray tracing

hrodmn:
I applied the same approach as we did in titiler-multidim (https://github.com/developmentseed/titiler-multidim/pull/103) and it is looking good - the traces/spans are all available in X-Ray and the logging statements are correlated which will make it possible to track performance of specific datasets, features, query patterns.

Here is a preview of an X-Ray trace with the correlated logs at the bottom:
image

### DIFF
--- a/titiler/cmr/main.py
+++ b/titiler/cmr/main.py
@@ -205,6 +205,6 @@ endpoints = Endpoints(
     title=settings.name,
     templates=templates,
     extensions=[TimeseriesExtension()],
-    enable_telemetry=True,
+    enable_telemetry=settings.telemetry_enabled,
 )
 app.include_router(endpoints.router)
--- a/titiler/cmr/main.py
+++ b/titiler/cmr/main.py
@@ -11,11 +11,15 @@ from starlette.templating import Jinja2Templates
 from titiler.cmr import __version__ as titiler_cmr_version
 from titiler.cmr.errors import DEFAULT_STATUS_CODES as CMR_STATUS_CODES
 from titiler.cmr.factory import Endpoints
+from titiler.cmr.logger import configure_logging
 from titiler.cmr.settings import ApiSettings, AuthSettings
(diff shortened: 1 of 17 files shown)


titiler/cmr/main.py
```diff
@@ -34,58 +31,6 @@
...
-# Only use custom JSON formatter locally - AWS Lambda handles JSON formatting automatically
-formatter: JSONFormatter | logging.Formatter
-if "AWS_EXECUTION_ENV" not in os.environ:
-    formatter = JSONFormatter()
-else:
-    formatter = logging.Formatter(
-        "%(asctime)s - %(name)s - %(filename)s:%(lineno)d - %(levelname)s - %(message)s"
-    )
-
-handler = logging.StreamHandler()
```
abarciauskas-bgse, line 77: Do any of the changes in this file impact logs showing up when running the application locally? Or running an instance without OpenTelemetry enabled?

abarciauskas-bgse (approved):
Thanks @hrodmn ! I would love it if we also got a review from @ividito since I don't have any experience with X-Ray or OpenTelemetry yet. But my only outstanding question is if this effects logging (and how) when open telemetry is not enabled.

ividito (approved):
This is really great! I'm super impressed by the Docker solution you put together here, since it also gets around one of the slowest parts of our lambda deployments.

hrodmn:
> But my only outstanding question is if this effects logging (and how) when open telemetry is not enabled.

Yeah, I need to clean up the logging configuration so that we get all of the logs in local deployments. My initial idea to move everything to handler.py basically eliminated local ogs. Thanks for the suggestion!

titiler/cmr/main.py
hrodmn: I updated the logging config to properly handle logs in the local context.

PR #93, opened 2025-11-03
DRY backend functions

abarciauskas-bgse:
This PR consolidates repeated code across the `#tile`, `#part` and `#feature` functions.


hrodmn (approved):
This is just a nit but can we put the new private methods at the top or the bottom of the class definition? I was confused at first when I saw them in the middle!

abarciauskas-bgse:
> This is just a nit but can we put the new private methods at the top or the bottom of the class definition? I was confused at first when I saw them in the middle!

Oh I definitely agree with that! https://github.com/developmentseed/titiler-cmr/pull/93/commits/4d7a49613170a370f41314ca1a84c684ff45efc6

PR #98, opened 2025-11-24
Feat/titiler cmr compatibility report

abarciauskas-bgse:
Questions:
- How many pie charts is too many? Are tables the only good alternative?
- What are the most important messages to get across in this report and should we highlight them more?
- Is any of the information uninteresting and unimportant, should it be removed?


docs/compatibility/compatible_datasets.csv
hrodmn, line 1: It doesn't have to happen in this PR but it would be awesome to add some kind of widget to the compatibility notebook that would allow users to browse this csv as a table, maybe with some kind of fuzzy searching.

hrodmn (commented):
@abarciauskas-bgse this is great - I just have a few questions/comments. The main things I am left wondering are:
- DAAC-specific _incompatibility_ - are there DAACS that are systematically less compatible?
- Maybe out-of-scope but just in case a data producer reads this and wants to make sure their dataset is compatible we could link out to a guide on best practices for producing compatible files
- consider hiding the code in the cells - I don't think we need to show the code in rendered docs page

abarciauskas-bgse:
@hrodmn I think I have addressed your comments - thank you for them! Let me know if you want to re-review.

hrodmn:
Thanks @abarciauskas-bgse - I tried running `uv run mkdocs serve` and got an error because `pyarrow` is missing from the environment. You will need to run `uv add --dev pyarrow` to get the docs to render properly.

The interactive table is great!

abarciauskas-bgse:
Nice catch! I added pyarrow to the dev dependencies

PR #100, opened 2025-11-25
feat: add permission boundary support

ividito:
Following up from #99 - our high security environment requires stacks to apply a permission boundary to all nodes. I think this is causing [deployment issues](https://github.com/NASA-IMPACT/veda-deploy/actions/runs/19651951520/job/56285764802) where the `ServiceRole` and `LogRetention` roles associated with the lambda function are missing this boundary.

### DIFF
--- a/infrastructure/aws/cdk/app.py
+++ b/infrastructure/aws/cdk/app.py
@@ -3,7 +3,7 @@
 import os
 from typing import Any, List, Optional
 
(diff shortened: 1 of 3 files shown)


botanical (approved):

hrodmn (approved):

hrodmn (commented):
@ividito the changes look good to me but you will need to apply the `ruff` formatting changes that pre-commit found!

PR #105, opened 2025-12-08
Avoid recording EDL token during testing.

chuckwondo:
## What I am changing

Removed accidentally recorded EDL access tokens from vcrpy yaml recording files, and updated vcrpy test config to avoid recording them in the future.

Also, a bit of related tidying:

- avoid pytest deprecation warning for
asyncio_default_fixture_loop_scope
- set UV_FROZEN and UV_NO_PROGRESS to 'true' in CI
- update some type annotations in tests

## How I did it

Scrubbed access token request/response entries from some vcrpy yaml files, and configured vcrpy configuration to skip recording requests to the EDL find_or_create_token endpoint.

## How you can test it

`uv run pytest`


hrodmn (approved):
Thanks for fixing this @chuckwondo!

PR #107, opened 2025-12-11
Remove deprecated FastAPI on_event decorator.

chuckwondo:
The `lifespan` function in `main.py` takes the place of the function removed in this commit, but it seems that when `lifespan` was added, the now-removed function was not removed, as it should have been.

(Also, tidied up some type annotations in the file where I removed the obsolete function.)

### DIFF
--- a/infrastructure/aws/lambda/handler.py
+++ b/infrastructure/aws/lambda/handler.py
@@ -4,14 +4,13 @@ import asyncio
 import logging
 import os
 import warnings
-from typing import Any, Dict
+from typing import Any
 
-import earthaccess
 from mangum import Mangum
+from mangum.types import LambdaContext, LambdaEvent
 
 from titiler.cmr.logger import configure_logging
 from titiler.cmr.main import app
-from titiler.cmr.settings import AuthSettings
 
 configure_logging()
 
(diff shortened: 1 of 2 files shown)


infrastructure/aws/lambda/handler.py
```diff
@@ -4,37 +4,24 @@
...
-    if auth_config.strategy == "environment":
-        app.state.cmr_auth = earthaccess.login(strategy="environment")
-    else:
-        app.state.cmr_auth = None
-
 
 handler = Mangum(
     app,
     lifespan="off",
-    api_gateway_base_path=None,
```
chuckwondo, line 37: This line didn't pass type checking because `api_gateway_base_path` is typed as `str`, not `str | None`. However, removing the line does not change the behavior because whether the argument is passed as `None` (before) or not passed at all (now), it defaults to `"/"`.

infrastructure/aws/lambda/handler.py
```diff
@@ -48,6 +35,6 @@ async def startup_event() -> None:
     loop.run_until_complete(app.router.startup())
 
 
-def lambda_handler(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
+def lambda_handler(event: LambdaEvent, context: LambdaContext) -> dict[str, Any]:
```
chuckwondo, line 38: FYI, this is actually redundant. This function can be eliminated, and the `handler` variable can simply be renamed to `lambda_handler`.

infrastructure/aws/lambda/handler.py
```diff
@@ -4,37 +4,24 @@
...
-auth_config = AuthSettings()
-
-
-@app.on_event("startup")
-async def startup_event() -> None:
-    """startup."""
-    if auth_config.strategy == "environment":
-        app.state.cmr_auth = earthaccess.login(strategy="environment")
-    else:
-        app.state.cmr_auth = None
```
chuckwondo, line 31: This is redundant (and obsolete) because this logic is handled in the `lifespan` function in `main.py`. This function should have been removed when `lifespan` was added because using `lifespan` replaces the now-deprecated use of the `on_event` decorator.

hrodmn (approved):

PR #109, opened 2026-01-08
Feat/reorg documentation

abarciauskas-bgse:
This PR restructures the documentation both to group notebooks under similar headings and to incorporate some of the notebooks that were previously in the Datacube Guide, but were about TiTiler-CMR.

The new structure (and some revised content) can be reviewed here: https://developmentseed.org/titiler-cmr/feat-reorg-documentation/

For future work (I can make these tickets if you agree @hrodmn )
- [titiler-cmr-statistics-benchmark.ipynb](https://github.com/developmentseed/datacube-guide/blob/f569eb44f9230f3f72ce85aed3af82699bc0caf7/docs/visualization/titiler/titiler-cmr/titiler-cmr-statistics-benchmark.ipynb) - I proposed that this go under benchmark/statistics but when re-running this I got zero statistics points returned and couldn't fix it after about an hour of finagling, so I removed it for now.
- [compatibility/tool.ipynb](https://github.com/developmentseed/titiler-cmr/blob/feat/reorg-documentation/docs/compatibility/tool.ipynb) - we probably want to modify this content and the compatibility tool further to say compatibility cannot be determined if 0 timesteps are returned.
- Update the time series api limits documentation given we are now using a container

Todo:
- [ ] delete branch from documentation's dropdown upon branch merge or delete


docs/packages/datacube_benchmark/pyproject.toml
hrodmn, line 1: I think we should remove this file and make sure all of these dependencies are present in the `docs` group in the project-level pyproject.toml. That way everything would be ready to run when we run `uv run --group docs mkdocs ...`

mkdocs.yml
```diff
@@ -25,17 +25,22 @@ extra:
...
+    - api/compatibility_api_example.ipynb
+    - api/tiling_api_xarray_backend_example.ipynb
+    - api/statistics_api_example.ipynb
+    - api/time_series_example.ipynb
+    - api/time_series_api_limits.ipynb
+    - api/rasterio_backend_example.ipynb
+    - api/sel_parameter_example.ipynb
+  - Benchmarking:
+    - benchmarking/tiles.ipynb
+    - benchmarking/hls_tiling.ipynb
```
hrodmn, line 38: I don't think the HLS notebook really belongs under the 'benchmarking' header. Maybe we could just rename `Benchmarking` to `Advanced` so interested readers can open it to find more without declaring an overall theme.

mkdocs.yml
```diff
@@ -25,17 +25,22 @@ extra:
...
-    - datasets/hls_tiling.ipynb
-  - Deployment:
-    - deployment/time_series_api_limits.ipynb 
-  - Development - Contributing: contributing.md
+  - How to use the API:
+    - api/compatibility_api_example.ipynb
+    - api/tiling_api_xarray_backend_example.ipynb
+    - api/statistics_api_example.ipynb
+    - api/time_series_example.ipynb
+    - api/time_series_api_limits.ipynb
```
hrodmn, line 33: Not blocking but we could just add the time series api limits notebook contents to the the main time series notebook. I'm not sure it needs to be a separate page.

docs/benchmarking/tiles.ipynb
hrodmn, line 1: the notebooks that use the datacube_benchmark python package are failing to render because it can't import the package :/

hrodmn (commented):
@abarciauskas-bgse thanks so much for the docs reorganization effort. I like the new structure! The main issue I see now is just that the notebooks that use the `datacube_benchmark` python package are not able to render properly because the package could not be imported when mkdocs rendered the site. I haven't dug in to see what is going on there - hopefully there is a simple fix!

docs/packages/datacube_benchmark/pyproject.toml
abarciauskas-bgse: I think it makes sense to leave it in as datacube_benchmark can then be installed as a standalone package, but let me know if you see it differently

mkdocs.yml
abarciauskas-bgse: I think they make sense as separate notebooks, since the limits notebook has explicit limits described

abarciauskas-bgse:
@hrodmn thanks for your review! I think I have addressed your comments, let me know what you think 🙏🏽

hrodmn:
@abarciauskas-bgse I think the changes all look good but the rendered docs site is exposing some problems in the new production deployment so I think we should switch the notebooks back to using the staging endpoint for now.

abarciauskas-bgse:
@hrodmn thank you for noticing that! I have replaced all instances of the production endpoint with the staging endpoint in https://github.com/developmentseed/titiler-cmr/pull/109/changes/a6d1cafda74f0b777d3cb0842faad913273909bd and looking at developmentseed.org/titiler-cmr/feat-reorg-documentation I believe the issues are resolved by using the staging endpoint.

PR #112, opened 2026-01-09
Add support for S3 temp creds + caching

chuckwondo:
I've added support for using s3 credentials, including caching them by provider.

I've added unit tests for this that leverages moto for mocking an s3 bucket and object.

I have not yet attempted to deploy to dev, but will pair up with Henry to do so, so we can do a manual integration test.

### DIFF
--- a/infrastructure/aws/lambda/Dockerfile
+++ b/infrastructure/aws/lambda/Dockerfile
@@ -3,10 +3,12 @@ ARG PYTHON_VERSION=3.12
 # Stage 1: OTEL
 # Download the OpenTelemetry layer
 # ref: https://github.com/athewsey/opentelemetry-lambda-container/blob/98069d5eb6d812ccd28d5c80e2f9d6c8a8c76fb9/python-example/lambda-function/Dockerfile
-FROM public.ecr.aws/lambda/python:${PYTHON_VERSION} as otel-builder
+FROM public.ecr.aws/lambda/python:${PYTHON_VERSION} AS otel-builder
+
 RUN <<EOF
 dnf install -y unzip wget
-wget https://github.com/aws-observability/aws-otel-python-instrumentation/releases/download/v0.12.1/layer.zip -O /tmp/layer.zip
+dnf clean all
+wget -q https://github.com/aws-observability/aws-otel-python-instrumentation/releases/download/v0.12.1/layer.zip -O /tmp/layer.zip
 mkdir -p /opt-builder
 unzip /tmp/layer.zip -d /opt-builder/
 EOF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,9 +21,9 @@ classifiers = [
     "Topic :: Scientific/Engineering :: GIS",
 ]
 dependencies = [
-    "titiler.core~=0.25.0",
-    "titiler.mosaic~=0.25.0",
-    "titiler.xarray~=0.25.0",
+    "titiler.core>=0.25.0,<0.26",
+    "titiler.mosaic>=0.25.0,<0.26",
+    "titiler.xarray>=0.25.0,<0.26",
     "aiobotocore>=2.24.0",
     "boto3>=1.39.0",
     "cftime~=1.6.4",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,9 +21,9 @@ classifiers = [
     "Topic :: Scientific/Engineering :: GIS",
 ]
 dependencies = [
-    "titiler.core>=0.23.1,<0.24",
-    "titiler.mosaic>=0.23.1,<0.24",
-    "titiler.xarray>=0.23.1,<0.24",
+    "titiler.core~=0.25.0",
+    "titiler.mosaic~=0.25.0",
+    "titiler.xarray~=0.25.0",
     "aiobotocore>=2.24.0",
     "boto3>=1.39.0",
     "cftime~=1.6.4",
@@ -42,6 +42,7 @@ dependencies = [
     "rioxarray~=0.13.4",
     "s3fs~=2025.9.0",
     "xarray~=2024.9.0",
+    "cachetools>=6.0.0",
 ]
 dynamic = ["version"]
 
@@ -81,6 +82,8 @@ dev = [
     "seaborn>=0.13.2",
     "uvicorn>=0.32.1",
     "pyarrow>=22.0.0",
+    "moto[s3,server]>=5.1.18",
+    "boto3-stubs[s3]>=1.42.24",
 ]
 deployment = [
     "aws-cdk-lib~=2.177.0",
@@ -132,7 +135,7 @@ known_third_party = [
 default_section = "THIRDPARTY"
 
 [tool.ruff]
-select = [
+lint.select = [
     "D1",  # pydocstyle errors
     "E",  # pycodestyle errors
     "W",  # pycodestyle warnings
@@ -140,7 +143,7 @@ select = [
     "C",  # flake8-comprehensions
     "B",  # flake8-bugbear
 ]
-ignore = [
+lint.ignore = [
     "E501",  # line too long, handled by black
     "B008",  # do not perform function calls in argument defaults
     "B905",  # ignore zip() without an explicit strict= parameter, only support with python >3.10
--- a/titiler/cmr/backend.py
+++ b/titiler/cmr/backend.py
@@ -321,25 +321,27 @@ class CMRBackend(BaseBackend):
         def _reader(asset: Asset, x: int, y: int, z: int, **kwargs: Any) -> ImageData:
             s3_credentials = self._get_s3_credentials(asset)
 
-            if isinstance(self.reader, type) and self.reader == Reader:
-                aws_session = self._create_aws_session(s3_credentials)
+            if any(
+                field.name == "opener_options" for field in attr.fields(self.reader)
+            ):
+                options = self._build_reader_options(s3_credentials)
 
-                with rasterio.Env(aws_session, **self.rasterio_env_kwargs):
+                with self.reader(
+                    asset["url"],
+                    tms=self.tms,  # type: ignore
+                    **options,
+                ) as src_dst:
+                    return src_dst.tile(x, y, z, **kwargs)
+            else:
+                with rasterio.Env(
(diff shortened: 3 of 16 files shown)


infrastructure/aws/lambda/Dockerfile
```diff
@@ -3,10 +3,12 @@ ARG PYTHON_VERSION=3.12
...
 # Download the OpenTelemetry layer
 # ref: https://github.com/athewsey/opentelemetry-lambda-container/blob/98069d5eb6d812ccd28d5c80e2f9d6c8a8c76fb9/python-example/lambda-function/Dockerfile
-FROM public.ecr.aws/lambda/python:${PYTHON_VERSION} as otel-builder
+FROM public.ecr.aws/lambda/python:${PYTHON_VERSION} AS otel-builder
+
 RUN <<EOF
 dnf install -y unzip wget
-wget https://github.com/aws-observability/aws-otel-python-instrumentation/releases/download/v0.12.1/layer.zip -O /tmp/layer.zip
+dnf clean all
+wget -q https://github.com/aws-observability/aws-otel-python-instrumentation/releases/download/v0.12.1/layer.zip -O /tmp/layer.zip
```
chuckwondo, line 11: Sorry, this was unrelated. Just my OCD kicking in.

titiler/cmr/backend.py
```diff
@@ -36,22 +50,31 @@
...
+    url: Union[str, Mapping[str, str]]
     provider: str
 
 
+class FindGranules(Protocol):
+    """Protocol for functions that search for granules in CMR."""
+
+    def __call__(self, count: int, **kwargs: Any) -> list[DataGranule]:
+        """Match signature of earthaccess.search_data."""
+        ...
```
chuckwondo, line 76: This is part of my additions for making it easy to unit test with a stub for earthaccess.search_data so we don't have to deal with more vcrpy recordings.

pyproject.toml
```diff
@@ -132,15 +135,15 @@ known_third_party = [
...
+lint.select = [
     "D1",  # pydocstyle errors
     "E",  # pycodestyle errors
     "W",  # pycodestyle warnings
     "F",  # flake8
     "C",  # flake8-comprehensions
     "B",  # flake8-bugbear
 ]
-ignore = [
+lint.ignore = [
```
chuckwondo, line 146: `select` and `ignore` are deprecated by ruff in favor of `lint.select` and `lint.ignore`.

pyproject.toml
```diff
@@ -21,9 +21,9 @@ classifiers = [
     "Topic :: Scientific/Engineering :: GIS",
 ]
 dependencies = [
-    "titiler.core>=0.23.1,<0.24",
-    "titiler.mosaic>=0.23.1,<0.24",
-    "titiler.xarray>=0.23.1,<0.24",
+    "titiler.core~=0.25.0",
+    "titiler.mosaic~=0.25.0",
+    "titiler.xarray~=0.25.0",
```
hrodmn: it's fine to upgrade to 0.25 but sometimes there are breaking changes between minor versions in titiler (it's up to 1.0 now so probably not likely to affect us here).
```suggestion
    "titiler.core>=0.25.0,<0.26",
    "titiler.mosaic>=0.25.0,<0.26",
    "titiler.xarray>=0.25.0,<0.26",
```

titiler/cmr/main.py
```diff
@@ -38,14 +44,50 @@
...
+    Returns:
+        A callable that accepts a provider identifier and returns temporary S3
+        credentials for that provider, via the ``auth`` object.
+    """
+
+    @cachetools.cached(
+        cachetools.TTLCache(maxsize=100, ttl=50 * 60),  # Expire in 50 minutes
+        condition=threading.Condition(),  # Prevent race conditions
+    )
+    @retry(5, HTTPException, 1)
```
hrodmn: This is fine but I just want to note that the cache will only persist as long as a lambda instance stays alive.

titiler/cmr/main.py
```diff
@@ -38,14 +44,50 @@
...
+    def get_s3_credentials_for_provider(provider: str) -> AWSCredentials:
+        # NOTE: Frustratingly, Auth.get_s3_credentials simply returns an empty
+        # dict if any sort of request fails, rather than raising an error.
+        # Therefore, we are forced to check the result and raise our own error
+        # if the result is empty.
+        if not (creds := auth.get_s3_credentials(provider=provider)):
+            # We cannot tell what the underlying exception was, since it was
+            # swallowed by earthaccess, so we're just making one up.
+            raise HTTPException(500)
```
hrodmn: A descriptive message would be nice

```suggestion
            raise HTTPException(500, "earthaccess failed to retrieve S3 credentials")
```

hrodmn (commented):
Thanks for the diligent work on the unit tests mocks and for persevering through the vcrpy morass! I like the way you refactored the application state `auth` to `get_s3_credentials` - it is much more clear what that is doing now.

These changes all look good to me. To get this to work in the Lambda context we will need to provide the EARTHDATA_USERNAME and EARTHDATA_PASSWORD environment variables so earthaccess can authenticate the S3 credentials request. I'm not sure which account we want to use for that, though. When we get to it we will want to define some GitHub Actions environment secrets and pass them along to the Lambda environment in the CDK script.

I am thinking about how to balance the need for flexibility in S3 access patterns with complexity in the application. Right now we will have to go all or nothing on EDL-based S3 credentials at deploy time. This is nice and simple, however, unless we solve the distributed cache for provider-specific credentials (thereby eliminating the need to request S3 credentials for every Lambda instance) I think it would not make sense to switch to the EDL method fully. 

Since the CMRBackend class now takes `get_s3_credentials` as an argument we could add a query parameter to the endpoints that could flip the switch for any request. This would make it so we could have a single titiler-cmr deployment that is capable of both authentication methods while we sort out a distributed cache for provider-specific credentials.

It would be fine to deploy this application as-is to the dev deployment with the EARTHDATA_* environment variables but I think it might be worth adding the query parameter authentication toggle so we don't wind up with a main branch that we don't want to deploy  to a production environment in the event of a need for a patch.

titiler/cmr/main.py
chuckwondo: Right. We want to look at a distributed cache approach after this.

chuckwondo:
> These changes all look good to me. To get this to work in the Lambda context we will need to provide the EARTHDATA_USERNAME and EARTHDATA_PASSWORD environment variables so earthaccess can authenticate the S3 credentials request. I'm not sure which account we want to use for that, though. When we get to it we will want to define some GitHub Actions environment secrets and pass them along to the Lambda environment in the CDK script.

Ah yes, I forgot to include github workflow changes. Shall I add to the PR for that? (regarding the EDL creds, any of us should be able to use our creds for the deployment -- there are no "service" accounts for EDL, so somebody needs to supply their creds)

> I am thinking about how to balance the need for flexibility in S3 access patterns with complexity in the application. Right now we will have to go all or nothing on EDL-based S3 credentials at deploy time. This is nice and simple, however, unless we solve the distributed cache for provider-specific credentials (thereby eliminating the need to request S3 credentials for every Lambda instance) I think it would not make sense to switch to the EDL method fully.
> 
> Since the CMRBackend class now takes `get_s3_credentials` as an argument we could add a query parameter to the endpoints that could flip the switch for any request. This would make it so we could have a single titiler-cmr deployment that is capable of both authentication methods while we sort out a distributed cache for provider-specific credentials.
> 
> It would be fine to deploy this application as-is to the dev deployment with the EARTHDATA_* environment variables but I think it might be worth adding the query parameter authentication toggle so we don't wind up with a main branch that we don't want to deploy to a production environment in the event of a need for a patch.

Since I'm still not particularly familiar with this app, can you clarify here? How would the query parameter be supplied (i.e., would the end-user need to know about this?)

How do you imagine the parameter would then be used? Some sort of flag to indicate whether to use "direct" or "external" access?

I'm happy to add it to this PR, if you think it would make sense.

titiler/cmr/backend.py
```diff
@@ -112,18 +137,15 @@ def _read(self) -> MosaicJSON:
...
-        if (
-            s3_auth_config.strategy == "environment"
-            and s3_auth_config.access == "direct"
-            and self.auth
-        ):
-            return aws_s3_credential(self.auth, asset["provider"])
-        else:
-            return None
+        return (
+            self.get_s3_credentials(provider)
```
abarciauskas-bgse: As an alternative to the parameter idea, could we whitelist which providers for which we DO NOT need to use EDL-based s3 credentials, at least until we solve distributed caching and simplify the codebase to JUST use EDL-based s3 credentials @hrodmn @chuckwondo ?

I think the parameter idea is fine too. I'm not too concerned about end-users having to know about it since the primary user would be applications using titiler-cmr, and the default can be no S3 credentials (that is, relying on the IAM role).

chuckwondo:
@abarciauskas-bgse and @hrodmn, I updated the README and CHANGELOG. Please let me know if the additions are sufficient for this PR.

titiler/cmr/backend.py
hrodmn: Yeah I think some kind of bucket-level whitelist could make sense since I think we know exactly which buckets are accessible via the veda-data-reader-dev role. We could check each asset href against the whitelisted buckets and only use EDL if the asset is in a bucket that is not whitelisted.

titiler/cmr/main.py
```diff
@@ -157,3 +194,8 @@ async def lifespan(app: FastAPI):
     enable_telemetry=settings.telemetry_enabled,
 )
 app.include_router(endpoints.router)
+
+app.state.auth = (auth := earthaccess.login(strategy="environment"))
```
hrodmn, line 198: Do you have any idea how long this takes to run? I wonder if adding something like this on each application startup could be expensive.

hrodmn (approved):
I think this all looks good - thanks for running down the earthaccess s3 credentials misdirection for the ASF provider!

titiler/cmr/main.py
chuckwondo: Good question. In my testing both locally and on MAAP ADE, I'm seeing roughly 2 seconds for login, fairly consistently. We also want to consider the time to _import_ of earthaccess. I've seen that vary from 1 to 3 seconds.

PR #116, opened 2026-01-14
Fix get_bbox_degrees for non-WGS84 CRSs.

### LINKED ISSUE #114: Function get_bbox_degrees returns incorrect miny value (2nd element in returned 4-tuple)
The function [get_bbox_degrees erroneously returns the maxx value for the miny value in the bbox it returns](https://github.com/developmentseed/titiler-cmr/blob/v0.3.1/titiler/cmr/utils.py#L89).

chuckwondo:
I corrected get_bbox_degrees and also added doctest unit tests for it.

Fixes #114

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -156,10 +156,23 @@ namespace_packages = true
 explicit_package_bases = true
 
 [tool.pytest.ini_options]
+minversion = "8.0"
 # Address PytestDeprecationWarning: The configuration option "asyncio_default_fixture_loop_scope" is unset.
 asyncio_default_fixture_loop_scope = "function"
 filterwarnings = [
     "ignore::rasterio.errors.NotGeoreferencedWarning",
 ]
 markers = "vcr: records network activity"
-addopts = "-Werror --cov=titiler.cmr --cov-report=term-missing --cov-report=xml -vv --benchmark-skip"
+addopts = [
+    "-Werror",
+    "--cov=titiler.cmr",
+    "--cov-report=term-missing",
+    "--cov-report=xml",
+    "-vv",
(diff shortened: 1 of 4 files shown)


hrodmn (commented):
@chuckwondo doctest looks cool, reminds me of the rust ecosystem for testing. Why use that instead of adding test in a test_utils.py file? I like the close-to-the-code approach but don't love the idea of formatting code in the docstrings.

chuckwondo:
> @chuckwondo doctest looks cool, reminds me of the rust ecosystem for testing. Why use that instead of adding test in a test_utils.py file? I like the close-to-the-code approach but don't love the idea of formatting code in the docstrings.

Putting such tests in the docs makes it convenient to see how to use the code when reading the docs. This works well for "pure" functions that use only the inputs passed and have no side-effects. It's very common amongst many of the packages we tend to use, such as numpy, (geo)pandas, and many others, where you'll see examples sections in the API docs. Here's a very simple example: https://numpy.org/doc/stable/reference/generated/numpy.ndarray.ndim.html

hrodmn (approved):
For a simple case like this I think it's fine - thanks for the fix!

chuckwondo:
> For a simple case like this I think it's fine - thanks for the fix!

Yep, anything with a bit of complexity would go in a regular text fixture, but for relatively simple, "pure" functions, doctests are my go-to.

I highly recommend using them for such cases because they are extremely illustrative. I find that they tend to be woefully underutilized, which is why I love the numpy, pandas, and geopandas docs (and other similarly documented libraries). The examples are clutch to understanding the functions, where extensive prose simply cannot effectively clarify alone.

PR #121, opened 2026-01-17
Update dependencies to address dependabot alerts

chuckwondo:
This PR updated all dependencies to the latest versions allowed by current dependency constraints, with the exception of urllib3. For now, we must remain on urllib 1.x, as 2.x has breaking changes with respect to gzip content decoding.

There were only 2 minor code changes required to keep unit tests passing, plus adjustments to vcrpy recordings due to upgrading earthaccess to 0.15.1.

This should address all current dependabot alerts, other than those for urllib3.


chuckwondo:
@hrodmn, I'm a bit confused as to why tests failed for this PR (see https://github.com/developmentseed/titiler-cmr/actions/runs/21084788268/job/60645955214), but did _not_ fail for your previous PR.

The point of failure was not code that was changed between the 2 PRs, so I would have expected your tests to fail for the same reason.

Given that you added new vars/secrets to the github `dev` environment, I would expect that the unit test workflow would need to be updated to run with the environment set to `dev` in order to pass, so I puzzled how your PR did _not_ fail tests.

chuckwondo:
@abarciauskas-bgse, @hrodmn, I believe this is in a reasonable state now.

I manually updated a handful of packages (via `uv lock -P <package>`) to address the majority (I think) of the dependabot alerts (see list below).

Also, I went back to using the `lifespan` function for calling `earthaccess.login` because logging in at the top level caused logins to occur unnecessarily, particularly during testing. It meant that unit tests would fail in the absence of any EARTHDATA env vars, which is not good, since there's no need for them during testing.

I've confirmed that the following updates don't appear to have broken anything (certainly not unit tests, but also the handful of urls we've been using to manually confirm functionality, including what we showed during PI demos -- of course, the nested rasterio env stuff is still not working, but we'll address that next).

```
    uv lock -P httpcore
    Updated h11 v0.14.0 -> v0.16.0
    Updated httpcore v1.0.7 -> v1.0.9
    
    uv lock -P fastapi
    Added annotated-doc v0.0.4
    Updated fastapi v0.115.5 -> v0.128.0
    uv lock -P starlette
    Updated starlette v0.41.3 -> v0.50.0
    
    uv lock -P setuptools
    Updated setuptools v75.6.0 -> v80.10.1
    
    uv lock -P tornado
    Updated tornado v6.4.2 -> v6.5.4
    
    uv lock -P aiohttp
    Updated aiohappyeyeballs v2.4.4 -> v2.6.1
    Updated aiohttp v3.11.9 -> v3.13.3
    Updated aiosignal v1.3.1 -> v1.4.0
    
    uv lock -P jinja2
    Updated jinja2 v3.1.4 -> v3.1.6
    
    uv lock -P requests
    Updated requests v2.32.3 -> v2.32.5
    
    uv lock -P filelock
    Updated filelock v3.16.1 -> v3.20.3
    
    uv lock -P virtualenv
    Updated virtualenv v20.28.0 -> v20.36.1
```

.github/workflows/ci.yml
```diff
@@ -57,4 +58,7 @@ jobs:
           uv run pre-commit run --all-files
 
       - name: Run tests
+        env:
+          EARTHDATA_USERNAME: ${{ vars.EARTHDATA_USERNAME }}
+          EARTHDATA_PASSWORD: ${{ secrets.EARTHDATA_PASSWORD }}
```
hrodmn: do we still need these changes or did the lifespan adjustment eliminate the need for earthdata variables for the unit tests.

titiler/cmr/main.py
```diff
@@ -39,6 +41,29 @@
...
+    elif auth is None:
+        # Set global auth instance only if not already set.  This allows for
+        # Lambda functions to avoid repeated logins during warm starts, also
+        # maintaining the s3 credential cache across warm starts.
+        app.state.auth = (auth := earthaccess.login(strategy="environment"))
+        app.state.get_s3_credentials = (
+            make_get_s3_credentials(auth) if auth_config.access == "direct" else None
+        )
+
+    yield
```
hrodmn, line 50: I think this is fine for testing and running locally, but we will need to handle it differently in the Lambda handler.

infrastructure/aws/lambda/handler.py
```diff
@@ -21,7 +21,7 @@
 
 handler = Mangum(
     app,
-    lifespan="off",
+    lifespan="on",
```
hrodmn: I believe we need to leave the lifespan `off` for the Lambda and define the startup event like this before building `handler`. I think this should make it so the login only happens on cold starts.

```python
@app.on_event("startup")
async def startup_event() -> None:
    app.state.auth = (auth := earthaccess.login(strategy="environment"))
    app.state.get_s3_credentials = (
        make_get_s3_credentials(auth) if auth_config.access == "direct" else None
    )

handler = Mangum(
    app,
    lifespan="off",
    text_mime_types=[
        "application/json",
        "application/javascript",
        "application/xml",
        "application/vnd.api+json",
    ],
)


if "AWS_EXECUTION_ENV" in os.environ:
    loop = asyncio.get_event_loop()
    loop.run_until_complete(app.router.startup())
```

hrodmn (changes requested):
Thanks for tackling the application startup situation. I have lost the thread on how that was related to the dependency upgrades but it's great to get this cleaned up regardless.

Based on [patterns from other eoapi-cdk services](https://github.com/developmentseed/eoapi-cdk/blob/ff4cde076f5f1a434fbf4e54eae7c0c37e7a600a/lib/titiler-pgstac-api/runtime/src/titiler_pgstac_api/handler.py#L81-L92), we need to use an explicit `startup` event (not lifespan) for the Lambda.

titiler/cmr/main.py
```diff
@@ -39,6 +41,33 @@
...
+
+@asynccontextmanager
+async def lifespan(app: FastAPI):
+    """FastAPI Lifespan."""
+
+    global auth
+
+    logger.info("Starting up")
+
+    if os.environ.get("PYTEST_VERSION"):
```
abarciauskas-bgse: would it be safer to explicitly set an environment variable (e.g. `ENV=test`) when running tests?

chuckwondo: When pytest is running, this is guaranteed to be set, so this won't work only if we plan to use a different test framework.

Either way, I'm not particularly fond of this logic for checking whether or not tests are running, so I'm certainly open to an alternative means of avoiding attempts to login to earthaccess during testing.

I suppose I could avoid the lifespan logic here and simply execute it only in the context of the lambda handler. What do you think about that? Since we don't have tests that test the lambda handler itself, this would avoid the need for such a check.

chuckwondo: Hmmm, I think that won't quite work either, as we do want the lifespan to run when we're using Docker locally, such that earthaccess login does occur.

infrastructure/aws/lambda/handler.py
chuckwondo: Yep, lifespan should be "off". However, on_event is deprecated and scheduled for removal in 1.x, so the lifespan async context manager is the recommended approach.

I'll set lifespan="off", but manually invoke startup, but not via `app.router.startup`, as that is tied directly to the use of `on_event("startup")`.

.dockerignore
```diff
@@ -1,7 +1,6 @@
...
-.mypy_cache
-.pytest_cache
-.ruff_cache
-
+*
+!titiler/
+!LICENSE
+!pyproject.toml
+!README.md
+!uv.lock
```
hrodmn, line 20: this approach does not play nice with AWS CDK :( https://github.com/developmentseed/titiler-multidim/pull/99#discussion_r2379838397

.github/workflows/ci.yml
chuckwondo: ah, yes, I should be able to remove these now

.dockerignore
chuckwondo: It does if you set `ignore_mode=IgnoreMode.DOCKER` in the call to `aws_lambda.DockerImageCode.from_image_asset`.

.github/workflows/ci.yml
chuckwondo: Removed

titiler/cmr/main.py
chuckwondo: I managed to eliminate this by changing the tests to avoid invoking the lifecycle events.

hrodmn (approved):

PR #131, opened 2026-02-04
fix: load S3 credentials and pass to openers in compatibility endpoint

### LINKED ISSUE #130: Implement EDL S3 access in /compatibility endpoint
The /compatibility endpoint is used to test compatibility of a concept id with the xarray and rasterio backends. It does not use the new S3 credential handling logic from recent EDL changes so it does not currently work with datasets from DAACs where we rely on EDL-based S3 access.

xarray backend:

https://github.com/developmentseed/titiler-cmr/blob/38e0d147543397adb0d8ae3d17c4bd2e8f749592/titiler/cmr/compatibility.py#L185-L207

rasterio backend:

https://github.com/developmentseed/titiler-cmr/blob/38e0d147543397adb0d8ae3d17c4bd2e8f749592/titiler/cmr/compatibility.py#L233-L258

I think we just need to run the S3 credential fetch process and apply it to the open/read operation for the sample granule in each of those functions.

cc @chuckwondo

hrodmn:
The compatibility endpoints do a very basic check to see if the first asset from the first granule be opened with either the xarray or the rasterio backends. This PR collects the S3 credentials via CMRBackend.get_s3_credentials and passes them onto the opening test accordingly.

resolves #130

### DIFF
--- a/tests/test_compatibility.py
+++ b/tests/test_compatibility.py
@@ -289,4 +289,4 @@ class TestConceptCompatibility:
             evaluate_concept_compatibility("C1234-TEST", mock_request, mock_auth)
 
         assert exc_info.value.status_code == 400
-        assert "cannot parse concept_id" in exc_info.value.detail
+        assert "Could not open a sample granule" in exc_info.value.detail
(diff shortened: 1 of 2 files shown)


chuckwondo (approved):
Great! Thanks Henry!

Issue #85, opened 2025-10-21
Determine solution for EDL-based access to DAAC buckets


abarciauskas-bgse:
Based on #84, determine + document what solution to use to support DAAC-based bucket access

chuckwondo:
@abarciauskas-bgse, this logic was implemented by #112 and #120.

However, there is no specific documentation for it (we simply used earthaccess and cached temporary s3 creds by s3 creds endpoint). Is there a place you would like some technical description added in order to consider this issue completed?

abarciauskas-bgse:
@chuckwondo we should probably update the documentation here: https://developmentseed.org/titiler-cmr/feat-reorg-documentation/#authentication-for-data-read-access

chuckwondo:
> [@chuckwondo](https://github.com/chuckwondo) we should probably update the documentation here: https://developmentseed.org/titiler-cmr/feat-reorg-documentation/#authentication-for-data-read-access

I think those docs looks complete from the user perspective.

Do you want me to add details about caching/auto-refresh of temp s3 creds there? That seems like it might be more of an implementation detail that's perhaps not relevant to the reader.

maxrjones:
> > [@chuckwondo](https://github.com/chuckwondo) we should probably update the documentation here: [developmentseed.org/titiler-cmr/feat-reorg-documentation#authentication-for-data-read-access](https://developmentseed.org/titiler-cmr/feat-reorg-documentation/#authentication-for-data-read-access)
> 
> I think those docs looks complete from the user perspective.
> 
> Do you want me to add details about caching/auto-refresh of temp s3 creds there? That seems like it might be more of an implementation detail that's perhaps not relevant to the reader.

fwiw I've become a fan of including "design notes" in the repo for recording these decisions, which could be [exposed in the docs](https://obspec-utils.readthedocs.io/en/latest/design/protocols/) or [not](https://github.com/pydata/xarray/tree/main/design_notes).

abarciauskas-bgse:
a design doc is a great idea!

@chuckwondo i was thinking of adding more details about how the credentials are being cached by provider there - enough information so that a new developer understand the mechanism at a high level. I also think the following warning may now be resolved?

```
At the moment, setting Earthdata credentials as environment variables does not work for the rasterio backend.
```

chuckwondo:
> a design doc is a great idea!
> 
[...]

PR #132, opened 2026-02-04
Describe temp s3 creds caching

chuckwondo:
@abarciauskas-bgse, how does this work for addressing this: https://github.com/developmentseed/titiler-cmr/issues/85#issuecomment-3849774757


hrodmn (approved):
Thanks for the clear explanation!

Issue #57, opened 2025-06-20
Supported datasets


abarciauskas-bgse:
It would be nice to keep track of supported datasets so we can test changes and releases.
 
* GPM IMERG (GES DISC): https://staging.openveda.cloud/api/titiler-cmr/tiles/WebMercatorQuad/0/0/0@3x?datetime=2020-10-04T00:00:00.000Z/2020-10-04T23:59:59.999Z&resampling=bilinear&variable=precipitation&colormap_name=gnbu&rescale=0,46&concept_id=C2723754864-GES_DISC&backend=xarray
* HLS (LP DAAC): https://staging.openveda.cloud/api/titiler-cmr/tiles/WGS1984Quad/9/295/144@1x?concept_id=C2021957657-LPCLOUD&datetime=2025-05-24T00%3A00%3A00Z%2F2025-05-24T23%3A59%3A59Z&post_process=swir&backend=rasterio&bands=B05&bands=B04&expression=1.5*((B05-B04)%2F(B05%2BB04%2B0.5))&rescale=-1,1&colormap_name=brbg&bands_regex=B[0-9][0-9] (note this is an example using the "expression" parameter.
* TROPESS (GES DISC): https://staging.openveda.cloud/api/titiler-cmr/tiles/WebMercatorQuad/0/0/0@3x?datetime=2021-12-01T00%3A00%3A00.000Z%2F2021-12-01T23%3A59%3A59.999Z&concept_id=C2837626477-GES_DISC&variable=o3&backend=xarray&colormap_name=reds&rescale=20%2C70&sel_method=nearest&reScale=0%2C5000&sel=lev%3D500&sel=time%3D2021-12-01T00%3A00%3A00.000Z

* MiCASA - GHG - coming soon

abarciauskas-bgse:
Minnie Wong sent me a list of 30 Worldview layers are using titiler-cmr, but they are all various indices of HLS:

Landsat L30 C2021957657-LPCLOUD

HLS_Customizable_Landsat https://go.nasa.gov/3TsL52o
HLS_False_Color_Landsat https://go.nasa.gov/4kMeQYd
HLS_False_Color_Vegetation_Landsat https://go.nasa.gov/447RsND
HLS_False_Color_Urban_Landsat https://go.nasa.gov/4jToHdp
HLS_Shortwave_Infrared_Landsat https://go.nasa.gov/3HJOYNP
HLS_NDVI_Landsat https://go.nasa.gov/43XViZG
[...]

PR #144, opened 2026-02-19
docs: add ADR for path-based backend routing for v1.0

### LINKED ISSUE #137: proposed v1.0.0 API changes
We need an ADR that describes the changes to the titiler-cmr for v1.0.0.

The main changes will be moving the backend logic (rasterio vs xarray) from a query parameter to a path parameter so that we can remove the backend switching control flows from the routing code.

hrodmn:
This ADR describes why I want to move the `backend` query parameter to a path parameter which will involve a breaking change to the API. At the end I describe a release plan which would translate nicely to a soft launch in the staging deployment.

resolves #137

### DIFF
--- a/design/2026-02-19-backend-path-routing.md
+++ b/design/2026-02-19-backend-path-routing.md
@@ -1,0 +1,131 @@
+# ADR: Backend-Specific URL Path Routing
+
+- **Status:** Proposed
+- **Date:** 2026-02-19
+- **Deciders:** @abarciauskas-bgse @chuckwondo @hrodmn
+
+---
+
+## Context
+
+titiler-cmr supports two raster backends for serving data from NASA CMR:
+
+- **rasterio** — for Cloud Optimized GeoTIFFs (COGs) and other GDAL-readable formats
+- **xarray** — for NetCDF, Zarr, and other array-oriented formats
+
+Supporting two backends with meaningfully different parameter sets within a single tiling API is inherently complex. The current design handles this by accepting a `backend` query parameter that selects the reader at runtime:
+
+```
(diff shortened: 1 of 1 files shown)


chuckwondo (approved):
Looks good to me. I think you've covered the bases well.

abarciauskas-bgse:
@hrodmn This is great! It clearly lays out major benefits to the codebase and end users. We should definitely move forward but I'll drop 2 additional thoughts:

1. Is the use of rasterio and xarray too specific? What if the backend we are using changes? For example we were talking about replacing rasterio with async-geotiff someday. Just a thought I don't think we should change the approach laid out here.
2. Communicating: when we have releases scheduled we should have plans for communication.
3. Monitoring and/or integration testing: should we have some planned work for integration testing and monitoring pre or post-v1.0.0 release?

abarciauskas-bgse:
@hrodmn should we merge this?

hrodmn:
Yes! I still need to address your comments though:

> 1. Is the use of rasterio and xarray too specific? What if the backend we are using changes? For example we were talking about replacing rasterio with async-geotiff someday. Just a thought I don't think we should change the approach laid out here.

Yeah I suppose we could think of a more generic name - maybe `cog`? I don't think we are using the rasterio backend for anything but COGs. I think for the ADR we can keep it as-is since the spirit of moving the reader backend into a path parameter will still apply.

> 2. Communicating: when we have releases scheduled we should have plans for communication.

Great point. Once we merge the changes from #145 I'll lay out a more detailed plan and get the word out to our titiler-cmr mailing list.

> 3. Monitoring and/or integration testing: should we have some planned work for integration testing and monitoring pre or post-v1.0.0 release?

I have tested the requests from #57 but will go through the other datasets that have been using titiler-cmr in the VEDA instances.

PR #145, opened 2026-03-04
feat: overhaul titiler-cmr api and backend

hrodmn:
This is a big one!

The original titiler-cmr backend and tiler factory classes re-implemented a lot of methods from the parent classes. This made it difficult to reason about the titiler-cmr-specific logic in this application. My goal in this PR is to use the latest versions of rio_tiler, titiler.core, titiler.mosaic, etc which should make most of the custom logic in this application unnecessary.

I have everything working locally for the new `/rasterio` and `/xarray` routes:

NISAR GCOV FrequencyB! ([map link](https://v4jec6i5c0.execute-api.us-west-2.amazonaws.com/xarray/WebMercatorQuad/map.html?collection_concept_id=C3622214170-ASF&bounding_box=-97,42,-85,48&variables=HHHH&variables=HVHV&rescale=-20,0&rescale=-30,-5&rescale=0,15&expression=10*np.log10(b1);10*np.log10(b2);10*np.log10(b1/b2)&group=%2Fscience%2FLSAR%2FGCOV%2Fgrids%2FfrequencyB&temporal=2026-01-01T00:00:00Z/2026-02-01T00:00:00Z))

image

HLS ([map link](https://v4jec6i5c0.execute-api.us-west-2.amazonaws.com/rasterio/WebMercatorQuad/map.html?bounding_box=-93,45,-89,49&collection_concept_id=C2021957657-LPCLOUD&sort_key=cloud_cover&temporal=2025-07-01T00:00:00Z/2025-07-30T23:59:59Z&assets_regex=B[0-9][0-9]&color_formula=Gamma%20RGB%203.5%20Saturation%201.2%20Sigmoidal%20RGB%2015%200.35&minzoom=7&assets=B04&assets=B03&assets=B02))
image

MUR SST ([map link](https://v4jec6i5c0.execute-api.us-west-2.amazonaws.com/xarray/WebMercatorQuad/map.html?datetime=2020-10-04T00:00:00.000Z/2020-10-04T23:59:59.999Z&resampling=bilinear&variable=precipitation&colormap_name=gnbu&rescale=0,46&concept_id=C2723754864-GES_DISC&backend=xarray))

image

### `CMRBackend`

The new structure is very similar to titiler-pgstac: the backend is a rio_tiler.mosaic backend with the `get_assets` method set up to query CMR using the Granule search query parameters that get passed directly onto the CMR API (using httpx client not earthaccess!). This class has a `dataset_reader` attribute that can be either a plain rio-tiler rasterio Reader or an XarrayReader. This is the part that mosaics images from multiple granule search results.

### `GranuleReader`
The GranuleReader is a MultiBaseReader (which can combine multiple assets into a single image). The GranuleReader.reader can be either a rasterio Reader or an XarrayReader.

### CMR Queries
I dropped earthaccess entirely in this PR. CMR API queries are handled with httpx and a set of new pydantic models to represent the search responses. Earthdata token and S3 credential fetching are also handled with httpx using the EARTHDATA_* environment variables.

I added the ability to `exitwhenfull` to stop fetching Granules once the search geometry is fully covered by the running list of returned granules.

I also added some more CMR Granule query parameters:
- cloud_cover
- sort_key (for sorting granules, really useful when paired with cloud_cover)
- granule_ur (useful for visualizing a specific granule)

## Tasks
- [x] Update backend.py
- [x] Update factory.py
- [x] Move `backend` from a query parameter to a path router
- [x] Implement timeseries changes
- [x] Implement compatibility changes
- [x] Implement S3 credential logic
- [ ] ~Add zarr support?~
- [x] Add routing for deprecated routes with deprecation warnings
- [x] Add aliases for new query parameters for backwards compatibility
- [x] Come up with plan for `expression` parameter breaking change (`B02/B03` -> `b1/b2`)
- [x] Update tests
- [x] Update docs
- [x] pin to titiler>=2.0,<3 and rio-tiler>=9.0,<10 when the new versions are released.

## Questions
- ~Since the Xarray backend does not need to be a MultiBaseReader (always a single asset) maybe I should define a GranuleRasterioReader(MultiBaseReader) and an GranuleXarrayReader(BaseReader)?~ I changed so the xarray backend uses a separate non-MultiBaseReader reader.
- The XarrayBackend feels quite slow in my local environment using the obspec-utils BlockStoreReader so I wonder if there is some way we could replicate the "background" caching behavior from fsspec that we have been using thus far. This might just be something specific about the structure of the dataset I was using. It seems to work fine for datasets like NISAR GCOV so maybe not worth worrying about for now.
- How closely should we mirror the CMR API? CMR uses `temporal` as the query parameter for datetime queries, but we use `datetime` in other applications (including titiler-cmr so far). I opted to keep the titiler-cmr API as true to the CMR API as possible in this go around because CMR API exceptions will be returned to the user and they are more helpful when the query parameters match.

cc @vincentsarago @chuckwondo @abarciauskas-bgse


docs/architecture.md
```diff
@@ -0,0 +1,116 @@
...
+`titiler-cmr` sits on top of `titiler.mosaic`, which provides a `MosaicTilerFactory` pattern. In this pattern a *backend* orchestrates multi-source search and tiling while a *reader* handles per-source data access. `CMRTilerFactory` extends `MosaicTilerFactory` with two backend variants that correspond to different data-access paths:
+
+- **Rasterio** — for granules that expose georeferenced raster assets (GeoTIFF, COG, etc.), accessed via `rasterio`.
+- **Xarray** — for granules that expose NetCDF/HDF5 files, opened via `xarray` and `open_dataset`.
+
+Each variant is registered as a separate router (e.g. `/rasterio/tiles/…` and `/xarray/tiles/…`) but shares the same `CMRBackend` for granule search. The reader class and its dependency are swapped to handle the format-specific opening logic.
+
+---
+
+## Class Hierarchy
```
abarciauskas-bgse, line 14: This architecture doc is great! But I am missing the context behind this class hierarchy diagram. How does `SpatialMixin` relate? What does each hierarchy represent and how do they relate to each other?

docs/architecture.md
```diff
@@ -0,0 +1,116 @@
...
+7. mosaic_reader merges N ImageData arrays using the configured pixel_selection method
+
+8. Post-process (algorithm) → render → Response(bytes)
+```
+
+For the **rasterio** backend, steps 2–6 are replaced by `MultiBaseGranuleReader`. It discovers the list of assets from the `Granule` at instantiation time and, for each asset, dispatches to a rasterio `Reader` (for COG/GeoTIFF) or `XarrayReader` (for NetCDF). `MultiBaseReader` handles iterating assets, merging per-asset results, and exposing them with index/expression filtering.
+
+---
+
+## Why Two Reader Base Classes?
```
abarciauskas-bgse, line 110: Again, missing a bit of background here -- where are the 2 reader base classes being introduced?

abarciauskas-bgse: If it easier to meet and discuss to clear up my understanding and then feed that back into this documentation that could be useful for me and hopefully others in the future.

abarciauskas-bgse:
This is exciting @hrodmn ! Since there are so many code changes are there ones we should focus on for review?

docs/architecture.md
hrodmn: I wanted to write down how the various reader/backend/factory classes arise from the upstream libraries. the SpatialMixin is the top of the chanin for rio-tiler's reader classes. I updated the diagram to be more clear, I think.

docs/architecture.md
hrodmn: They are defined in titiler/cmr/reader.py, there is one for rasterio backend and one for xarray backend. Maybe 'base' class is the wrong term.

titiler/cmr/expression.py
hrodmn, line 1: I needed to add this so that we could lazily evaluate array math when evaluating expressions in the XarrayReader. numexpr was eagerly loading the entire array. This is different than the titiler.xarray XarrayReader which does this operation at the ImageData level where eagerly running the calculation is fine. In this case we run the operation on the actual DataArray before the data are cropped to a tile.

titiler/cmr/factory.py
```diff
@@ -1,1063 +1,220 @@
...
-                        image.band_names = [xarray_ds_params.variable]
+###############################################################################
+# Standalone granules router — backend-independent CMR granule metadata queries
 
-                    stats = image.statistics(
-                        **stats_params.as_dict(),
-                        hist_options=histogram_params.as_dict(),
-                        coverage=coverage_array,
-                    )
+granules_router = APIRouter()
```
hrodmn, line 90: I opted to set up this top-level router instead of having redundant pairs of routes (e.g. /xarray/.../granules and /rasterio/.../granules). These functions don't open files so they don't need the reader to be involved.

titiler/cmr/legacy.py
hrodmn, line 1: This is where I have the logic for forwarding requests to pre v1 endpoints onto the new routes (e.g. `/tiles/WebMercatorQuad/x/y/z?backend=rasterio` -> `/rasterio/tiles/WebMercatorQuad/x/y/z?...`

titiler/cmr/models.py
hrodmn, line 1: This module has all of the pydantic models for the CMR search functionality. We could try to spin this out (with query.py) into earthaccess or possibly a new package for interacting with CMR.

titiler/cmr/models.py
```diff
@@ -1,912 +1,585 @@
...
-        str,
-        Field(
-            description="Reference to one coordinate reference system (CRS) as WKT string",
-            examples=[
-                'GEOGCS["GCS_WGS_1984",DATUM["D_WGS_1984",SPHEROID["WGS_1984",6378137,298.257223563]],PRIMEM["Greenwich",0],UNIT["Degree",0.017453292519943295]]',
-            ],
-        ),
-    ]
 
+class GranuleSearch(BaseModel):
```
hrodmn, line 141: We could add any of the granule search parameters from the [CMR API docs](https://cmr.earthdata.nasa.gov/search/site/docs/search/api.html#granule-search-by-parameters)

titiler/cmr/query.py
```diff
@@ -0,0 +1,214 @@
...
+    if search_shape is None or granule.geometry is None:
+        return False, covered
+    covered = covered.union(shapely.geometry.shape(granule.geometry))
+    check = (
+        search_shape.buffer(coverage_tolerance) if coverage_tolerance else search_shape
+    )
+    return check.covered_by(covered), covered
+
+
+def get_granules(
```
hrodmn, line 113: This is the heart of the entire application. Takes a GranuleSearch and returns granules.

hrodmn:
@chuckwondo @abarciauskas-bgse @vincentsarago I am finally done working on this branch.

I realize that such a massive PR is really difficult to review so please let me know if you want any help getting context for the changes. I am feeling really good about the state of things but if you want any changes or clarification let me know.

Here is a rough outline of all of the changes:

1. Upgrade to rio-tiler>=9,<10 and titiler>=2,<3
2. Use the new MosaicTilerFactory class from titiler.mosaic to define a very thin CMRTilerFactory (no more re-defining all of the factory class methods)
3. Update the CMRBackend class to mimic PgSTACBackend: all it does is run granule queries for tile/bbox/point requests
4. Handle CMR query and S3 credential internally rather than with earthaccess
   - uses pydantic models and httpx to create a structured query interface (I would love to spin this out of titiler-cmr someday)
   - allows flexible CMR queries by surfacing many of the CMR Granule query parameters directly to the titiler-cmr API (granule_ur, cloud_cover, attribute, etc)
   - S3 credential and Earthdata token credential handling based on the NasaEarthdataCredentialProvider from obstore
5. Use obstore + obspec-utils to access data for the xarray reader (no more fsspec)

I recommend taking a tour of the docs site that I published for the branch:
https://developmentseed.org/titiler-cmr/feat-restructure

There is one new notebook that goes strategies and considerations for rendering data from collections that is organized into orbital swaths: https://developmentseed.org/titiler-cmr/feat-restructure/api/visualizing_orbital_swath_data

docs/architecture.md
```diff
@@ -0,0 +1,188 @@
+# Architecture
+
+## Overview
+
+`titiler-cmr` sits on top of `titiler.mosaic`, which provides a `MosaicTilerFactory` pattern. In this pattern a *backend* orchestrates multi-source search and tiling while a *reader* handles per-source data access. `CMRTilerFactory` extends `MosaicTilerFactory` with two backend variants that correspond to different data-access paths:
```
abarciauskas-bgse:
```suggestion
`titiler-cmr` sits on top of `titiler.mosaic`, which provides a `MosaicTilerFactory` pattern. TiTiler's object-oriented design includes 2 foundational class types: *backends* and *readers*. In this pattern a *backend* orchestrates multi-source search and tiling while a *reader* handles per-source data access. `CMRTilerFactory` extends `MosaicTilerFactory` with two backend variants that correspond to different data-access paths:
```

docs/architecture.md
abarciauskas-bgse: Ah ok, perhaps you could just add some introductory text: 

> The diagram below illustrates how classes in TiTiler-CMR relate to the rest of the TiTiler class hierarchy. At the top of the TiTiler class hierarchy is `rio_tiler.io.base.SpatialMixin`.

docs/architecture.md
```diff
@@ -0,0 +1,188 @@
...
+   is created and stored at `app.state.earthdata_token_provider`. The provider lazily fetches and
+   refreshes the bearer token from `https://urs.earthdata.nasa.gov/api/users/find_or_create_token`
+   on demand rather than eagerly at startup.
+2. If `EARTHDATA_S3_DIRECT_ACCESS=true`, a `GetS3Credentials(token_provider)` instance is
+   constructed (taking the provider so it can call `token_provider()` and pick up refreshed tokens)
+   and stored at `app.state.get_s3_credentials`.
+
+`GetS3Credentials` maintains a `TTLCache(maxsize=100, ttl=50m)` internally. Calling it with
+an endpoint URL either constructs a new `EarthdataS3CredentialProvider` or returns the cached
+instance for that endpoint. The 50-minute TTL covers the expected lifetime of a Lambda execution
```
abarciauskas-bgse: Is 50 minutes the limit for a lambda runtime which may be reused across executions? Lambda executions have a limit of 15 minutes so just wanted to check my understanding of this time limit.

docs/migration.md
```diff
@@ -0,0 +1,66 @@
...
+```
+
+Update to the new paths when possible to avoid the extra round-trip. Once migrated,
+remove the `?backend=` parameter — it has no effect on the new `/rasterio/` or `/xarray/`
+paths.
+
+## New parameters
+
+- **`orbit_number`** — Filter granule search results to a specific orbital pass number.
+- **`skipcovered`** — Skip granules whose spatial footprint is already fully covered by
```
abarciauskas-bgse: is this a boolean parameter?

titiler/cmr/expression.py
abarciauskas-bgse: is this new functionality? This may work towards resolving [this issue](https://github.com/US-GHG-Center/ghgc-architecture/issues/824) which identified that lack of some math operations for the MiCASA data as NetCDFs through TiTiler-CMR was a blocker for using TiTiler-CMR for that dataset.

docs/architecture.md
chuckwondo: No, 50 minutes is simply 10 minutes short of 60 minutes, which is how long the temp creds last.  The 10-minute buffer has nothing to do with Lambda, and only to do with making sure creds get refreshed in advance of their actual expiry so that they don't get used too close to their actual expiry to avoid having them expire in the middle of using them.

docs/migration.md
hrodmn: yes it is boolean. it is ported over from pgstac which will not return two items that have the same exact geometry which is useful to limit the number of assets that rio-tiler has to consider when rendering a tile.

docs/architecture.md
hrodmn: the 50 minute timeout will be relevant when we work on a distributed cache for S3 credentials. it might be relevant if we use provisioned concurrency for these functions but right now we aren't.

chuckwondo: I think the confusing wording is this, however: "The 50-minute TTL covers the expected lifetime of a Lambda execution"

It is confusing because the ttl is completely unrelated to the execution environment, but the wording above seems to indicate that they are related somehow.

The ttl is strictly related to how long the temporary creds last (1 hour), regardless of the execution environment. Whether running locally, in a Lambda, in an EC2 instance, or wherever, the ttl will still be 50 minutes (unless we choose a different timeout buffer for some reason, but whatever the reason, it would be unrelated to the environment).

It is also unrelated to how we're caching, whether a local cache or a distributed cache.

docs/migration.md
```diff
@@ -0,0 +1,66 @@
...
+   tile, image, statistics, and timeseries path. Choose the prefix that matches the data
+   access method your collection uses. This was previously handled with the `backend` query
+   parameter.
+
+   > The `/timeseries`, `/bbox/{bbox}/granules`, `/point/{lon},{lat}/granules`, and
+   > `/tiles/{tms}/{z}/{x}/{y}/granules` endpoints are backend-agnostic and keep the same
+   > path — no prefix needed.
+
+2. **`concept_id` renamed to `collection_concept_id`.** The query parameter identifying a
+   CMR collection has been renamed to match CMR's own terminology.
```
chuckwondo, line 15: Just to add a bit of clarity, CMR supports using `concept_id` to specify a collection's concept ID for granule queries.  Collection concept ID's start with a `"G"`, so CMR recognizes that when passing such an ID via the `concept_id` parameter in a granule query.

However, I like that you have changed this to `collection_concept_id`, even though it is unnecessary, because it provides clarity (and the CMR also supports it, in addition to `concept_id` in a granule query).  It makes it unambiguous that we specifically want a collection concept ID.

infrastructure/aws/lambda/Dockerfile
```diff
@@ -50,15 +50,17 @@ rm -rf numpy/doc/ bin/ geos_license Misc/
...
-find . -name 'LICENSE*' -delete
+find . -name 'LICENSE*' -print0 | xargs -0 rm -rf
 find . -name 'README*' -delete
 find . -name '*.md' -delete
 # Strip debug symbols from shared libraries (preserve numpy.libs)
 find . -type f -name '*.so*' -not -path "*/numpy.libs/*" -exec strip --strip-unneeded {} \; 2>/dev/null || true
-# Remove *.dist-info directories
-find /deps -mindepth 1 -maxdepth 1 -type d -name '*.dist-info' -exec rm -rf {} \;
-# Remove unused botocore services (we are using only s3, so that stays)
-find /deps/botocore/data -mindepth 1 -maxdepth 1 -not -path "*/s3" -exec rm -rf {} \;
```
chuckwondo, line 61: I think you might have mistakenly removed this bit.

titiler/cmr/models.py
```diff
@@ -1,912 +1,585 @@
...
+from shapely.geometry import shape
+
+from titiler.cmr.errors import S3CredentialsEndpointMissing
+
+# ---------------------------------------------------------------------------
+# Query parameter type aliases
+# ---------------------------------------------------------------------------
+
+ConceptID = Annotated[
+    str | None,
```
chuckwondo, line 29: Let's remove all `| None` from all of the annotated types.  This is generally not a desirable way to allow for `None` values because it completely prevents us from using the annotated type in cases where we do _not_ want to permit `None` values.

Instead, the `| None` should be added only at the locations where we want to allow it (i.e., where this annotated type is _used_, not here, where it is _defined_).

Not only does this change allow us to use the type in places where we do not want to allow for `None` values, but it also aids in readability at usage points.  Where we want to allow for `None`, we must specifiy it as `ConceptID | None`, making it clear that we allow for `None`, but conversely, if `| None` is part of the definition, a usage of `ConceptID` as a type annotation masks the fact that `| None` is part of the definition, which is generally counterintuitive and unexpected.

titiler/cmr/models.py
```diff
@@ -1,912 +1,585 @@
...
+    """CMR granule search parameters."""
 
-class CRSRef(BaseModel):
-    """CRS from referenceSystem."""
+    model_config = ConfigDict(populate_by_name=True)
 
-    referenceSystem: Dict[str, Any] = Field(
-        ...,
-        description="A reference system data structure as defined in the MD_ReferenceSystem of the ISO 19115",
+    collection_concept_id: ConceptID = Field(
```
chuckwondo, line 146: Corresponding to my suggestion to remove `| None` from the _definition_ of `ConceptID`, we want to add it here, like so:

```suggestion
    collection_concept_id: ConceptID | None = Field(
```

We want to make the changes to the definitions and usages across all of the annotated types (not only for `GranuleID`).

titiler/cmr/models.py
```diff
@@ -1,912 +1,585 @@
...
-class Temporal(BaseModel):
-    """Temporal Extent model.
+        CMR interprets a bare datetime (e.g. "2024-01-01T00:00:00Z") as an
+        open-ended range starting at that date. Convert it to a closed interval
+        ("2024-01-01T00:00:00Z/2024-01-01T00:00:00Z") so only granules for
+        that exact instant are matched.
+        """
+        if v is not None and "/" not in v:
+            return f"{v}/{v}"
+        return v
```
chuckwondo, line 173: I'm on the fence with this.  Given that you're trying to expose the same param names that CMR uses (which I agree with, because we also pass back CMR error messages to the user, so using the same param names avoid mismatching the names in the error messages, as you've mentioned), this change seems (very minimally) counter to that thinking.  I suggest we simply drop this and pass it straight through to the CMR so the user gets the same behavior as if hitting the CMR directly.

This rather might be addressed by opening a CMR documentation issue indicating that they should document this behavior on their side. This is simply an ommission in their docs about how a bare datetime is treated.

titiler/cmr/models.py
chuckwondo: While I'd like to introduce something similar into earthaccess, I am not at all a fanboy of Pydantic.  I find it bloated, convoluted, and not as performant as I'd like, not even v2.  Msgspec would be my preference (leaner, cleaner, and more performant), but given that Pydantic is already in our dependency hierarchy, this is fine here, but not my recommendation for earthaccess (which is already bloated and needs trimming, or a least modularization via refactoring into "extras").

docs/migration.md
abarciauskas-bgse: @chuckwondo did you mean to say 

>Collection concept ID's start with a "C", so CMR recognizes the `concept_id` as the collection to use in searching for granules

Its a bit confusing in the documentation - I think you can use `concept_id` or `collection_concept_id` to identify the collection to search on in a granules query. [This documentation](https://cmr.earthdata.nasa.gov/search/site/docs/search/api.html#granule-search-by-parameters) only provides examples using `collection_concept_id` but [Search after](https://cmr.earthdata.nasa.gov/search/site/docs/search/api.html#search-after) uses examples with `concept_id`

chuckwondo: Sorry, yes, I meant "C", not "G".

Yes, for granule search, you can specify a collection concept ID as the value of either the concept_id or collection_concept_id param.  If using concept_id to specify the collection concept ID, since collection concept IDs start with a "C", the CMR will do the right thing.  That is, it checks the starting character of the value of the concept_id param to disambiguate the type of object you're specifying the concept ID for, whereas using collection_concept_id requires no disambiguation.

chuckwondo (changes requested):
Thanks again for this overhaul @hrodmn!

I still have more review to do, but didn't want to leave my comments so far to remain pending, so I'm submitting a review now so you can see them, but will continue reviewing and submit more as I go.

docs/migration.md
hrodmn: I like the explicit `collection_concept_id`. I had to change it because when you search with `granule_ur` for a specific granule you are required to provide `collection_concept_id` with it. 

We could also keep `concept_id` and a user could just provide a granule's own G-... concept id instead of the collection's C-... concept id and a datetime range.

I checked this request and it appears to work so the `granule_ur` addition was not exactly adding a new feature, we just hadn't used titiler-cmr with a granule concept id yet:
https://staging.openveda.cloud/api/titiler-cmr/tiles/WebMercatorQuad/0/0/0@3x?resampling=bilinear&variable=precipitation&colormap_name=gnbu&rescale=0,46&concept_id=G3993405945-GES_DISC&backend=xarray

titiler/cmr/expression.py
hrodmn: Until now the xarray backend did not have `expression` capability which makes it easy to apply a transformation to the raw values. The other new feature in the xarray backend is to make it possible to select multiple variables for visualization so that should unblock the `NBE = NEE + FIRE + FUEL` visualization case.

We can convert the NPP values from kg C/m2/s to g C/m2/day with `expression=NPP*86400000` then apply `rescale=-4,4`: [preview image](https://v4jec6i5c0.execute-api.us-west-2.amazonaws.com/xarray/bbox/-180,-90,180,90.png?collection_concept_id=C3273638632-GES_DISC&datetime=2024-12-31T00:00:00Z&variables=NPP&expression=86400000*NPP&rescale=-4,4&colormap_name=rdylgn)

The linear combination to get NBE = NEE + FIRE + FUEL works too. You have to urlencode the expression in your browser but most http clients will do that for you automatically:
https://v4jec6i5c0.execute-api.us-west-2.amazonaws.com/xarray/bbox/-180,-90,180,90.png?collection_concept_id=C3273638632-GES_DISC&datetime=2024-12-31T00:00:00Z&variables=NEE&variables=FIRE&variables=FUEL&expression=86400000*(NEE%2BFIRE%2BFUEL)&rescale=-4,4&colormap_name=rdylgn

titiler/cmr/models.py
hrodmn: Yeah I would like to be consistent with CMR API as much as possible, but it presents a great opportunity for a silent failure case if we unwittingly let titiler-cmr forward the open-ended datetime interval to CMR. It will return a result successfully but it might not be the result you were expecting! It might also crash titiler-cmr if it returns too many granules, but we don't have an explicit error case for this.

docs/migration.md
chuckwondo: No, I agree with your decision to go with collection_concept_id.  I was just adding some pedantic clarification around what the CMR handles.

titiler/cmr/models.py
chuckwondo: Unfortunately, this is an ambiguous situation because it is not documented in the CMR (again, raising this with CMR support is something I'd like to see), and the "correction" logic being applied here assumes the user has made a mistake.  They could very well be attempting to leverage the undocumented CMR behavior (that they are perhaps aware of).

Thus, this logic might very well produce unexpected results for the user.  That is, they may very well expect this to behave as the lower bound of an open temporal range.

Therefore, given the ambiguity, I suggest we forbid such an ambiguous temporal value and raise an error indicating that it is an ambiguous temporal range because we cannot determine whether this is intended to be a single point in time, the lower bound of an open range, or the upper bound of an open range.

Generally speaking, in the face of ambiguity, I prefer not to assume we know what the user is thinking (or not thinking) because doing something based on a wrong assumption can be equally as bad as doing nothing.

chuckwondo: In this case, we can utterly avoid making a poor assumption by completely forbidding the ambiguity.

hrodmn: I like that solution. Just require a closed interval.

infrastructure/aws/lambda/Dockerfile
hrodmn: It was not a mistake because we are not installing botocore in our dependencies anymore during the builder stage. However we can probably still clean them out of the Lambda runtime directory if they are still left around in the final stage which copies the dependencies from the builder stage into the final stage.

chuckwondo: Ah, I see what you did in pyproject.toml to refactor the dependencies.  Makes sense.

pyproject.toml
```diff
@@ -53,37 +48,39 @@ uvicorn = [
 
 lambda = [
     "mangum==0.19.0",
-    "aiobotocore>=2.24.0,<2.24.2"
 ]
 
 [dependency-groups]
 dev = [
+    "boto3-stubs[s3]>=1.42.24",
```
chuckwondo, line 56: I like the deps reorg, but let's keep boto3 as an explicit dependency since we do import it directly.  I know boto3-stubs includes boto3, but it's much safer to explicitly list dependencies used directly rather than depend on transitive dependence.

titiler/cmr/backend.py
```diff
@@ -1,215 +1,161 @@
...
-    Literal,
-    Optional,
-    Protocol,
-    Tuple,
-    Type,
-    TypedDict,
-    Union,
-)
+from collections.abc import Callable
+from typing import Any, Type, cast
```
chuckwondo: You can use `type` instead of `Type`, and that should deal with the last of the deprecated typing cases.

titiler/cmr/backend.py
```diff
@@ -1,215 +1,161 @@
...
-    crs: CRS = attr.ib(default=WGS84_CRS)
-    geographic_crs: CRS = attr.ib(default=WGS84_CRS)
-
-    # The reader is read-only (outside init)
-    mosaic_def: MosaicJSON = attr.ib(init=False)
+    """Mosaic backend for CMR granule search."""
 
-    input: str = attr.ib("CMR", init=False)
+    # CMR search parameters
+    input: GranuleSearch = attr.ib()  # type: ignore[assignment]
```
chuckwondo: How about naming this more specifically, such as `query` or `granule_query`.  Seeing `self.input` in the code is ambiguous.

titiler/cmr/credentials.py
```diff
@@ -0,0 +1,185 @@
...
+        self._expires_at: datetime | None = None
+        self._on_refresh: list[Callable[[], None]] = []
+
+    def __call__(self) -> str:
+        """Return the current token, refreshing if near expiry."""
+        with self._lock:
+            if not self._is_valid():
+                self._fetch()
+                for cb in self._on_refresh:
+                    cb()
```
chuckwondo: We can create a "fast path" by using double-checked locking to avoid unnecessary locking:

```suggestion
        if not self._is_valid():
            with self._lock:
                if not self._is_valid():
                    self._fetch()
                    for cb in self._on_refresh:
                        cb()

```

titiler/cmr/credentials.py
```diff
@@ -0,0 +1,185 @@
...
+        self.__dict__.update(state)
+        self._lock = threading.Lock()
+
+    def __call__(self) -> "S3Credential":
+        """Return cached credentials, refreshing if near expiry."""
+        with self._lock:
+            if not self._is_valid():
+                self._cached = self._fetch()
+            assert self._cached is not None
+            return self._cached
```
chuckwondo, line 155: We can create a "fast path" by using double-checked locking to avoid unnecessary locking:

```suggestion
        if not self._is_valid():
            with self._lock:
                if not self._is_valid():
                    self._cached = self._fetch()

        assert self._cached is not None
        return self._cached
```

titiler/cmr/credentials.py
```diff
@@ -0,0 +1,185 @@
...
+        return EarthdataS3CredentialProvider(endpoint, self._token_provider())
+
+    def clear(self) -> None:
+        """Clear the credential provider cache."""
+        with self._lock:
+            self._cache.clear()
+        logger.info("S3 credential provider cache cleared after token refresh")
+
+
+class EarthdataS3CredentialProvider:
```
chuckwondo, line 121: Any reason you're not using `obstore.auth.NasaEarthdataCredentialProvider`? It already provides this logic, allowing for use of an auth token and automatic refreshing.

titiler/cmr/utils.py
```diff
@@ -37,7 +37,16 @@ def retry(
...
+    """Decorator that retries a wrapped function on specified exceptions.
+
+    If all retry attempts are exhausted, the function is called one final time
+    without catching exceptions, so any error propagates to the caller.
+
+    Args:
+        tries: Number of guarded retry attempts before the final unguarded call.
+        exceptions: Exception type or sequence of types to catch and retry on.
+        delay: Seconds to wait between retry attempts.
+    """
```
chuckwondo, line 49: Thanks for improving the docstring!

titiler/cmr/utils.py
```diff
@@ -67,7 +76,25 @@ def _parse_date(date: str) -> datetime:
...
+
+    Returns:
+        Tuple of ``(datetime_, start, end)``.  For a bare datetime,
+        ``datetime_`` is set and both interval fields are ``None``.  For an
+        interval, ``start`` and/or ``end`` are set (each may be ``None`` for
+        open bounds) and ``datetime_`` is ``None``.
+
+    Raises:
+        InvalidDatetime: If any component cannot be parsed.
+    """
```
chuckwondo, line 97: Thanks for improving the docstring!

titiler/cmr/timeseries.py
```diff
@@ -241,8 +277,22 @@ def build_request_urls(
...
+        method: HTTP method — either ``"GET"`` or ``"POST"``.
+        **kwargs: Additional arguments forwarded to the underlying httpx method
+            (e.g. ``json``, ``timeout``).
+
+    Returns:
+        The httpx Response object.
+
+    Raises:
+        ValueError: If ``method`` is not ``"GET"`` or ``"POST"``.
+    """
```
chuckwondo, line 293: 👍

titiler/cmr/settings.py
```diff
@@ -20,6 +32,7 @@ class ApiSettings(BaseSettings):
     time_series_image_max_total_size: float = 1e8
     telemetry_enabled: bool = False
     debug: bool = False
+    cmr_timeout: float = 10.0
```
chuckwondo, line 35: 👍

titiler/cmr/query.py
```diff
@@ -0,0 +1,214 @@
...
+    search_params: GranuleSearch,
+    client: Client,
+    geometry: Geometry | None = None,
+    page_size: int = 10,
+    limit: int = 100,
```
chuckwondo: I don't know what's sensible for the app, but if a user does not override these defaults, then we could end up making 10 requests (100 granules / 10 per page).

This might be a tricky balance to optimize performance, but I suspect we could bump the page size up a bit to reduce the number of requests.

titiler/cmr/query.py
```diff
@@ -0,0 +1,214 @@
...
+                granule, search_shape, covered, coverage_tolerance
+            )
+            if done:
+                logger.info("Search geometry fully covered, stopping early")
+                return
+
+        if not (cmr_search_after := response.headers.get("cmr-search-after")):
+            break
+
+        headers["cmr-search-after"] = cmr_search_after
```
chuckwondo, line 194: I'd like to see the CMR fetching logic cleanly separated from the coverage logic, but we can defer that, as this is already a massive PR, and you have things working.  Just noting for some future refactoring.

Specifically, we could do the CMR querying/pagination completely isolated in a separate function, which would return an Iterator/Generator.  Then the coverage logic can be perform completely separately, and simply halt iteration of the granules as necessary.  There would be no change in mem/perf from what you have here, just a clean/clear separation of logic to aid readability and debugging (if necessary).

chuckwondo (commented):
@hrodmn, well done on such a massive lift!

Most of my comments deal with minor code adjustments.

I've scanned the bulk of the logic changes (although not with a fine-toothed comb, due to the volume of code) and didn't see any logic errors, as far as my limited understanding goes at least.

Hopefully we can cover the comments and questions on our call with @abarciauskas-bgse tomorrow and you can relatively quickly address any agreed upon changes.

Thanks again!

titiler/cmr/credentials.py
hrodmn: I thought there might be merit to requesting the EDL token at application startup, then only requesting S3 credentials at the moment they are needed (in the Reader classes). However after another look I think this might be over-complicating it. I'll simplify the approach to re-use NasaEarthdataCredentialProvider.

hrodmn: I thought there might be merit to requesting the EDL token at application startup, then only requesting S3 credentials at the moment they are needed (in the Reader classes). However after another look I think this might be over-complicating it. I'll simplify the approach to re-use NasaEarthdataCredentialProvider.

titiler/cmr/backend.py
hrodmn: I agree that it is somewhat ambiguous but we are inheriting from a class where `input` is a required argument so it would be more work to rename it to something more descriptive.

titiler/cmr/credentials.py
hrodmn: I refactored it to use NasaEarthdataCredentialProvider. The EarthdataTokenProvider remains for handling https auth bearer tokens for https requests which we only use when running the application locally.

titiler/cmr/models.py
hrodmn: I tried this but FastAPI needs the Annotated types to contain None to mark them as optional query parameters.

titiler/cmr/backend.py
chuckwondo: Ah, sure, makes sense.

docs/migration.md
hrodmn: Actually, I would like to leave the temporal range handling as-is for now and track it in a follow-up issue since I want to be careful about more breaking changes here https://github.com/developmentseed/titiler-cmr/issues/146

titiler/cmr/query.py
hrodmn: I think it's worth putting a larger page size here since fetching 100 granules should be really fast compared to the overhead of making 10 requests. However I don't think we are going to use more than 10 granules very many tile requests. For now I'll set the page to 20.

titiler/cmr/models.py
chuckwondo: Yes, but it shouldn't be here.

See my corresponding comment to this one, where I add the `| None` bit to the model that uses this, specifically where the `GranuleSearch` model defines `collection_concept_id`.

At that spot, you need this:

```python
collection_concept_id: ConceptID | None = Field(default = None, ...)
```

The same applies to the rest of the Annotated items.  Drop the `| None` in the Annotation _definition_ and add it where the annotation is _used_.

abarciauskas-bgse (approved):

chuckwondo (approved):
Kudos @hrodmn! Massive, much-needed overhaul.  I know you want to do more testing before we release, but approving so you're not blocked.

PR #148, opened 2026-03-30
feat: convert Lambda from container to zip

### LINKED ISSUE #139: Convert Lambda from container image to .zip archive
Once #138 is implemented, we may very well be under the 250 MiB limit for Lambda .zip archives, and thus should be able to switch our Lambda over to a .zip archive package rather than a container image.

If not, we would likely be extremely close, and implementing #64 and #74 should likely get us over the hump. Even if #138 buys us enough space, implementing these other 2 issues should shave off another several MiBs, giving us a bit more wiggle room.

hrodmn:
This PR unlocks zip archive Lambda functions! It is deployed to the dev endpoint and it's much snappier, especially on cold starts. Here is a [map.html with HLS tiles](https://v4jec6i5c0.execute-api.us-west-2.amazonaws.com/rasterio/WebMercatorQuad/map.html?bounding_box=-93,45,-89,49&collection_concept_id=C2021957657-LPCLOUD&sort_key=cloud_cover&temporal=2025-09-01T00:00:00Z/2025-09-30T23:59:59Z&assets_regex=B%5B0-9%5D%5B0-9%5D&color_formula=Gamma%20RGB%203.5%20Saturation%201.2%20Sigmoidal%20RGB%2015%200.35&minzoom=7&assets=B04&assets=B03&assets=B02) that I have been using for testing.

The key change is installing some system libraries up front including a fairly minimal GDAL installation, installing python packages from source so they use the shared system libraries, and then ensuring the system libraries are included in the final `/asset` directory. Before this change rasterio, pillow, pyproj, shapely, etc each installed their own copies of various system libraries. The CDK docker build takes longer than it used to so I want to explore some kind of build cache later on.

The other changes include:

* enable SnapStart for minimizing cold start times
* export OTEL + xray traces directly from the Lambda handler instead of running a sidecar process. this was tricky, I am surprised it is not easier to set up considering there are already libraries which supposedly handle it but I had to do a fair bit of fiddling with signing, finding xray-trace-id headers, etc.

I added a few notes (helped by Claude) to summarize the architectural changes in docs/architecture.md


infrastructure/aws/cdk/app.py
```diff
@@ -71,14 +69,6 @@ def __init__(
             iam.PermissionsBoundary.of(self).apply(permissions_boundary_policy)
             Aspects.of(self).add(PermissionsBoundaryAspect(permissions_boundary_policy))
 
-        iam_reader_role = None
-        if app_settings.role_arn:
-            iam_reader_role = iam.Role.from_role_arn(
-                self,
-                "veda-reader-dev-role",
-                role_arn=app_settings.role_arn,
-            )
```
hrodmn, line 80: We don't need to provide a special role anymore now that we rely on EDL for S3 access.

uv.lock
hrodmn, line 1: just a few security patch updates and the new telemetry dependencies

infrastructure/aws/lambda/Dockerfile
```diff
@@ -1,88 +1,191 @@
...
+find . -type f -name '*.so*' -exec strip --strip-unneeded {} \;
 EOF
 
-CMD ["handler.lambda_handler"]
+# Bundle GDAL data files (schemas, SRS definitions) needed by the GDAL drivers
+COPY --from=gdal-builder /opt/gdal/share/gdal /asset/gdal_data
+
+COPY infrastructure/aws/lambda/handler.py /asset/handler.py
+
+CMD ["echo", "hello world"]
```
vincentsarago, line 191: 🤩 this is really cool 💪

infrastructure/aws/lambda/handler.py
```diff
@@ -14,8 +23,171 @@
...
+# ── SnapStart pre-warming ──────────────────────────────────────────────────────
+# Build the Starlette/OTEL middleware stack now so the first post-restore request
+# doesn't pay the assembly cost.
+app.middleware_stack = app.build_middleware_stack()
+
+# Open a rasterio.Env and leave it open (do not call __exit__).  This calls
+# gdal.AllRegister() and keeps the reference count at 1 so the GDAL driver
+# manager is not torn down.  gdal.AllRegister() is not re-entrant-safe so
+# request handlers that use their own rasterio.Env() push on top of this one
+# rather than re-initializing from scratch.
```
vincentsarago, line 182: 👀

infrastructure/aws/lambda/handler.py
```diff
@@ -14,8 +23,171 @@
...
+_gdal_env = rasterio.Env()
+_gdal_env.__enter__()
+
+# Warm the PROJ CRS cache so the first CRS lookup after restore hits the
+# in-memory cache rather than the SQLite database.
+pyproj.CRS.from_epsg(4326)
```
vincentsarago, line 188: 🔥

chuckwondo (approved):
Phenomenal @hrodmn! It looks like the `/asset` dir in the container is down to 193M!

hrodmn:
> It looks like the /asset dir in the container is down to 193M!

Yes! I had it down to 182M but had to do something about libcurl that bumped it up to 193.

PR #163, opened 2026-04-10
docs: point docs at production endpoint (with override option)

hrodmn:
Now that we have a production deployment we should point the docs to that it instead of the staging deployment!

I also added the option to provide a different api endpoint in case we want to build the docs against a different deployment (staging, dev, etc) when we run the workflow manually.

I built the docs site on this branch to preview the change:
https://developmentseed.org/titiler-cmr/docs-update-endpoint/


chuckwondo (approved):
Thanks @hrodmn!

PR #166, opened 2026-04-13
Fix zizmor errors

chuckwondo:


.github/actions/cdk-deploy/action.yml
```diff
@@ -36,9 +37,9 @@ runs:
       run: npm install -g aws-cdk@2
 
     - name: Install uv
-      uses: astral-sh/setup-uv@v7
+      uses: astral-sh/setup-uv@cec208311dfd045dd5311c1add060b2062131d57 # v8.0.0
       with:
-        version: "0.5.*" 
+        version: "0.5.*"
```
hrodmn: while we are in here we could upgrade to `0.11.*`

hrodmn (dismissed):

.github/actions/cdk-deploy/action.yml
chuckwondo: Good call.

chuckwondo: I bumped it in benchmark.yml as well.

hrodmn (approved):

PR #169, opened 2026-04-21
Add ECCO Sea Ice Velocity dataset documentation

hanzila1:
## Summary
Adds comprehensive documentation for the ECCO L4 Sea Ice Velocity dataset (C1990404790-POCLOUD), demonstrating all TiTiler-CMR capabilities with real working examples.

## Changes
- New notebook: `docs/api/ecco_sea_ice_velocity_example.ipynb`
- End-to-end tests: `tests/test_ecco_sea_ice_velocity.py` 
- Updated navigation: `mkdocs.yml`

## Dataset Details
- Monthly Arctic sea ice velocity data (1992-2018)
- Two velocity components: SIeice (eastward), SInice (northward)
- 0.5° global grid, -0.4 to +0.4 m/s range

## Demo Coverage
✅ Dataset compatibility inspection  
✅ Interactive tile visualization with folium  
✅ Regional statistics computation  
✅ Multi-temporal analysis  
✅ Variable comparison (eastward vs northward)  
✅ PNG tile generation and display  

## Testing
All endpoints validated with VCR-recorded tests covering:
- `/compatibility` - metadata inspection
- `/tiles/WebMercatorQuad` - PNG tile generation  
- `/xarray/statistics` - regional statistics
- `/timeseries` - temporal analysis

## Future Enhancement
As suggested by @hrodmn , this could be extended with ParticleLayer visualization for animated sea ice motion vectors (https://docs.mapbox.com/mapbox-gl-js/example/raster-particle-layer/).

Happy to make any adjustments based on teams feedback!

PR #172, opened 2026-04-28
feat: add MVT endpoint for granule footprints (closes #153)

### LINKED ISSUE #153: Granules as vector tiles
It is not efficient to return granules as geojson for visualization purposes. It would be very cool to add some vector tile endpoints that return granules as vector tiles with all the titiler-cmr query parameters. Maybe in this application or somewhere else.

hanzila1:
## Summary

Closes #153

The existing granule endpoints return GeoJSON which loads all data at once — inefficient for map visualization at scale. This PR adds `f=mvt` support to the existing tile endpoint so clients can fetch only the granules visible in the current viewport as binary vector tiles.

### Backend changes
- Added `f=mvt` option to `GET /tiles/{tms}/{z}/{x}/{y}/granules`
- Granule geometries are clipped to tile bounds using `shapely` before encoding
- Binary MVT response encoded via `mapbox-vector-tile`
- All existing CMR query parameters (`collection_concept_id`, `temporal`, etc.) supported
- Added `mapbox-vector-tile>=2.2.0` to dependencies

### Frontend demo
Added `docs/api/granule_vector_tiles_demo/` — a standalone MapLibre GL JS demo that:
- Fetches only tiles visible in the current viewport
- Shows live stats: visible granules, tiles fetched, zoom level
- Supports any CMR collection universally
- Links each granule popup to its CMR metadata

Tested with ECCO Sea Ice Velocity and MUR SST (Global) collections.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,6 +29,7 @@ dependencies = [
     "h5py>=3.15.1",
     "httpx>=0.28.1",
     "isodate>=0.7.2",
+    "mapbox-vector-tile>=2.2.0",
     "obspec-utils>=0.9.0",
     "pillow>=12.1.1",
     "psutil>=7.2.2",
--- a/titiler/cmr/factory.py
+++ b/titiler/cmr/factory.py
@@ -1,11 +1,13 @@
 """titiler.cmr.factory: router factories."""
 
 import logging
-from typing import Annotated, Callable, Literal
+from typing import Annotated, Callable, Literal, Union
 
 import morecantile
 from attrs import define, field
 from fastapi import APIRouter, Depends, Path, Query
+from fastapi.responses import Response
+from shapely.geometry import box, shape
 from rio_tiler.constants import WGS84_CRS
 from titiler.core.dependencies import (
     DatasetParams as RasterioDatasetParams,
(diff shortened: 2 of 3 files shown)


hanzila1:
Here's a quick demo of the implementation working end to end:

https://github.com/user-attachments/assets/9463e97c-63f9-4806-b4d0-03eb292944e7

The demo shows:
- Granule footprints rendered as MVT - only tiles in the current viewport are fetched from NASA CMR
- Live tile counter incrementing as MapLibre requests each tile
- Clicking a granule shows its ID, temporal range, and links to the raw CMR metadata

Tested with ECCO Sea Ice Velocity (C1990404790-POCLOUD) and MUR SST Global (C1996881146-POCLOUD).

hanzila1:
@hrodmn 
just wanted to flag that I've updated this PR significantly based on your feedback pattern from #169.

The frontend demo has been extracted into its own standalone repository (same approach as the ECCO PR, and will be sharing deployed version soon, your review needed)
This PR now only touches 3 files - the backend MVT endpoint logic in factory.py, the mapbox-vector-tile dependency in pyproject.toml, and the updated uv.lock.

The demo video in the original comment still shows the full end-to-end flow (latest upcoming demo will show additional features to interact with granule metdata, instead of complex json). 

Happy to address any feedback on the backend implementation approach, particularly around whether the shapely geometry clipping logic in assets_for_tile looks correct to you, or if you'd prefer a different approach?

Regards

PR #175, opened 2026-05-05
fix: avoid duplicating GranuleSearch params in timeseries requests

### LINKED ISSUE #174: some query parameters getting duplicated in /timeseries sub request urls
Any GranuleSearch query parameters in a time series request will get duplicated which is a problem for query parameters with an array type.

I discovered this when trying to do some HLS time series statistics requests.

e.g. `/rasterio/timeseries/statistics?collection_concept_id=C123456&sort_key=cloud_cover&temporal=2025-01-01T00:00:00Z/2025-01-05T00:00:00Z&temporal_mode=point&step=P1D` will yield statistics requests like 
`/rasterio/statistics?collection_concept_id=C123456&sort_key=cloud_cover&temporal=2025-01-01T00:00:00Z/2025-01-02T00:00:00Z&collection_concept_id=C123456&sort_key=cloud_cover` (note the duplicated query parameter values). Duplicates for `collection_concept_id` are tolerated but since `sort_key` can be an array it becomes a problem.

hrodmn:
This did not come up in my testing during the overhaul but it is a failure case. I also noticed that the resulting bad requests returned an empty set of statistics results which is not the expected behavior! Now any 400/500 errors from the lower-level requests will get raised to the timeseries requester.

resolves #174


chuckwondo (approved):
LGTM!

PR #177, opened 2026-05-08
chore: replace dynamic versioning with release-please

hrodmn:
Instead of managing versions by hand let's use release-please to maintain the changelog and versioning. We lose the hash-based version diff from uv-dynamic-versioning but save some manual effort on long term maintenance.


chuckwondo (approved):
Excellent! Thanks @hrodmn.

PR #186, opened 2026-06-03
feat: add group inspection to compatibility endpoint

### LINKED ISSUE #185: Add group hints to the `/compatibility` endpoint for hierarchical xarray datasets
## Problem

The `/compatibility` endpoint is supposed to help users understand how to call TiTiler-CMR for a collection. That works well for flat NetCDF/HDF5 datasets because the response surfaces variables, dimensions, coordinates, and example links.

For hierarchical HDF5 and NetCDF datasets, that is not enough. Some collections are only usable through the xarray backend when the caller also provides a `group` query parameter, but `/compatibility` does not currently expose that requirement or any candidate group paths.

NISAR collections are a concrete example. A response like the one below can report `backend: "xarray"` while still omitting the information needed to actually make a successful tiling request:

```json
{
  "concept_id": "C3622214170-ASF",
  "backend": "xarray",
  "datetime": [{"RangeDateTimes": [{"BeginningDateTime": "2025-10-01T00:00:00.000Z"}], "EndsAtPresentFlag": true}],
  "variables": {},
  "dimensions": {},
  "coordinates": {},
  "example_assets": "s3://.../NISAR_L2_PR_GCOV_...h5",
  "sample_asset_raster_info": null,
  "links": [
    {"rel": "tilejson", "href": ".../rasterio/..."},
    {"rel": "map", "href": ".../rasterio/..."},
    {"rel": "tile", "href": ".../rasterio/..."}
  ]
}
```

Today, a user has to know out of band that they should try a `group` value.

## Proposal [Updated]

Extend `/compatibility` so that, for xarray-compatible collections backed by hierarchical files, it returns a list of groups that are compatible with titiler.

The endpoint should stay concise. It does not need to become a full explorer for every group and every variable. It should just surface enough information for a caller to discover that:

- the sample asset contains hierarchical groups
- the collection likely requires `group=...` for xarray requests
- there are one or more candidate group paths to try

Suggested additions to the xarray compatibility response:

- `compatible_groups: list[str]`

We would also add an optional `group` query parameter to the /compatibility endpoint so that users can test the compatibility of each group returned in `compatible_groups`.

## User value

- Makes `/compatibility` actually actionable for hierarchical datasets
- Reduces trial-and-error for HDF5 and grouped NetCDF collections
- Prevents the misleading case where a collection appears xarray-compatible but still cannot be used without undocumented extra parameters
- Keeps the discovery workflow in the endpoint that already exists to help users craft requests

## Acceptance criteria

- [ ] When the sample xarray asset has no groups, `/compatibility` behavior stays effectively unchanged
- [ ] When the sample xarray asset has hierarchical groups, `/compatibility` indicates that group selection is relevant
- [ ] The response includes one or more candidate group paths that a caller can pass as `group`
- [ ] The response shape remains lightweight and does not attempt to serialize full per-group dataset metadata
- [ ] Tests cover grouped and non-grouped xarray compatibility responses
- [ ] Relevant docs/examples are updated if the response contract changes

## Implementation notes

- `titiler/cmr/compatibility.py` already opens a sample asset and extracts top-level xarray metadata. This seems like the natural place to add lightweight group inspection.
- `titiler/cmr/reader.py` already supports `group` in `open_dataset(...)` and `XarrayGranuleReader`, so this is a discovery/documentation gap more than a net-new reader capability.
- The compatibility docs describe `/compatibility` as an endpoint that helps users craft the right request parameters, which makes group hints a good conceptual fit.
- Keep this lightweight. A dedicated xarray inspection endpoint could come later if richer hierarchy browsing is needed.

## Open questions

- What is the safest way to identify candidate groups without opening and inspecting too much of a large HDF5 hierarchy?
- Should the endpoint return only group paths, or also a tiny amount of per-group summary information such as variable counts?
- If multiple groups look valid, should `/compatibility` pick a `default_group`, or should it avoid guessing?
- Should xarray links include `group` only when a default is confidently known, to avoid encoding a misleading recommendation?

## Alternatives considered

- Add a dedicated `/xarray/...` discovery endpoint instead. Better for rich inspection, but heavier than the immediate need.
- Do nothing and rely on external docs or notebooks. That keeps the API simple but preserves a frustrating discovery gap for grouped datasets.

## Validation

- [ ] Add or update unit tests around `evaluate_xarray_compatibility` and `evaluate_concept_compatibility`
- [ ] Verify the response against a known grouped collection such as NISAR
- [ ] Verify that a flat xarray collection still returns the current lightweight metadata shape

hrodmn:
This adds the ability to inspect hierarchically organized HDF5 files. Now if the sample granule has nested groups it will check the dimensions of each group and return a list of `compatible_groups` that should be accessible for titiler-cmr. It won't inspect variables across all of the possible groups but there is a new `group` query parameter on the compatibility endpoint which can be used to learn about the variables within a single group (once you know what the groups are).

resolves #185

### DIFF
--- a/tests/test_compatibility.py
+++ b/tests/test_compatibility.py
@@ -8,6 +8,9 @@ from fastapi import HTTPException
 
 from titiler.cmr.compatibility import (
     CompatibilityResponse,
+    _candidate_group_paths,
+    _dataset_dim_scale_names,
+    _group_has_spatial_dims,
     evaluate_concept_compatibility,
     evaluate_rasterio_compatibility,
     evaluate_xarray_compatibility,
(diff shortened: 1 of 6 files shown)


chuckwondo (changes requested):
@hrodmn, the main logic looks reasonable, but let's avoid as much of the magic mocking as possible.  It is relatively easy to construct in-memory h5 files, so let's do that instead of using all of the _Fake* classes.  Too much mocking can give us a false sense of test coverage.

hrodmn:
@chuckwondo good call, I replaced the mocks with some actual HDF5 fixtures.

tests/test_compatibility.py
```diff
@@ -64,95 +107,215 @@ def _make_granule(external_href="https://example.com/file.nc") -> Granule:
...
 
+def _make_xarray_dataset(
+    data_vars: dict[str, tuple[tuple[str, ...], np.ndarray]],
+    coords: dict[str, np.ndarray] | None = None,
+) -> xr.Dataset:
+    """Create a small in-memory xarray dataset for tests."""
+    return xr.Dataset(
+        data_vars={name: (dims, values) for name, (dims, values) in data_vars.items()},
+        coords=coords,
+    )
```
chuckwondo: This seems redundant.  All calls to `_make_xarray_dataset` can be replaced with a direct call to `xr.Dataset`.

titiler/cmr/compatibility.py
```diff
@@ -157,16 +168,185 @@ def extract_xarray_metadata(
...
+
+            file_handle.visititems(visitor)
+
+            if spatial_group_paths:
+                return spatial_group_paths
+            if spatial_metadata_group_paths:
+                return spatial_metadata_group_paths
+            return all_group_paths
+    finally:
+        reader.close()
```
chuckwondo:
```suggestion

    all_group_paths: list[str] = []
    spatial_group_paths: list[str] = []
    spatial_metadata_group_paths: list[str] = []

    def visitor(name: str, obj: Any) -> None:
        if not (
            name
            and isinstance(obj, h5py.Group)
            and any(isinstance(child, h5py.Dataset) for child in obj.values())
        ):
            return

        all_group_paths.append(name)

        if not _group_has_spatial_dims(obj):
            return

        if "/metadata/" in f"/{name.strip('/')}/":
            spatial_metadata_group_paths.append(name)
        else:
            spatial_group_paths.append(name)

    with (
        _make_blockstore_reader(src_path, auth_token, credential_provider) as reader,
        h5py.File(reader, "r") as file_handle,
    ):
        file_handle.visititems(visitor)

    return spatial_group_paths or spatial_metadata_group_paths or all_group_paths
```

chuckwondo (dismissed):
Thanks @hrodmn.  I made 2 minor suggestions, but feel free to ignore; not blocking.

chuckwondo (approved):
Nice work @hrodmn!

PR #192, opened 2026-06-09
chore(deps): upgrade aws-cdk-lib to upgrade Node.js runtime

### LINKED ISSUE #191: LogRetention Lambda functions on outdated Node.js 20.x runtime
The CDK stack deploys a Lambda function to manage log retention. These functions use a Node.js runtime that is controlled by the aws-cdk-lib version that we use to deploy the stack. To upgrade the Node.js runtime to 22.x we will need to upgrade to a newer version of aws-cdk-lib. Right now we are pinned to 2.177.x in uv.lock but the latest version is 2.258.x

cc @botanical

hrodmn:
I checked locally and by upgrading to aws-cdk-lib==2.258.1 the runtime should get upgraded to Node.js 22.x

resolves #191

### DIFF
--- a/.github/actions/cdk-deploy/action.yml
+++ b/.github/actions/cdk-deploy/action.yml
@@ -32,10 +32,6 @@ runs:
         path: ~/.npm
         key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
 
-    - name: Install AWS CDK
-      shell: bash
-      run: npm install -g aws-cdk@2
-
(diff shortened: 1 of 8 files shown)


botanical (approved):
Pulled locally to test these changes using test-cdk-synth and it works for me! Do you think it's worth adding an `.nvmrc` file?

hrodmn:
> Pulled locally to test these changes using test-cdk-synth and it works for me! Do you think it's worth adding an .nvmrc file?

Sure we could do that!

PR #193, opened 2026-06-09
chore: record httpx2 traffic with vcr-py

### LINKED ISSUE #194: after httpx2 switch some pytest tests are becoming flaky
I am not 100% sure about the source of the problem but I have been experiencing intermittent pytest failures in a few cases today. I had a bot do some digging and it appears to be related to the vcr-py fixtures not properly catching requests from httpx2.

See this CI run for an example of what we are seeing: https://github.com/developmentseed/titiler-cmr/actions/runs/27223525881/job/80390112598?pr=192

hrodmn:
These new fixtures were generated by GPT-5.4!

Resolves #194

### DIFF
--- a/tests/cassettes/test_xarray_backend/test_timeseries_gif.yaml
+++ b/tests/cassettes/test_xarray_backend/test_timeseries_gif.yaml
@@ -1,0 +1,153 @@
+interactions:
+- request:
+    body: ''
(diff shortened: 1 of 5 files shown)


chuckwondo (approved):
Thanks @hrodmn!

PR #197, opened 2026-06-17
ci: set default branch to main in deploy-dev action workflow

hrodmn:
I was confused when trying to run the deploy-dev action with the default settings because I didn't notice that it was still trying to deploy the non-existent `develop` branch!


chuckwondo (approved):
Ugh! Good catch!

PR #202, opened 2026-06-18
fix: install titiler 2.0.5 to avoid missing httpx dependency

hrodmn:
Our dependency upgrades from #198 dropped `httpx` as a (non-dev) dependency which broke titiler/cmr/reader.py when it tried importing `httpx` in `titiler.xarray.io`. This was a bug (https://github.com/developmentseed/titiler/issues/1432) that is now fixed in titiler 2.0.5!

We can fully eliminate httpx as a dependency once titiler 2.1.0 comes out but for now we will need to keep httpx and httpx2 installed.


chuckwondo (approved):
Good catch @hrodmn!

PR #205, opened 2026-06-27
fix: add granule_ur to compatibility params and response

### LINKED ISSUE #204: Add granule_ur to the /compatibility endpoint
## Goal

Allow `/compatibility` callers to pass an optional `granule_ur` query parameter so compatibility checks can run against a known granule instead of always using the first sampled granule. Return the sampled `granule_ur` in the response so callers know which granule was evaluated.

## Proposal

- Add `granule_ur` as an optional `/compatibility` query parameter.
- Use it when building the CMR `GranuleSearch` for the sample granule.
- Include the sampled granule's `granule_ur` in the `/compatibility` response.
- Preserve the current behavior when `granule_ur` is omitted.

## Acceptance criteria

- [ ] `/compatibility?collection_concept_id=...&granule_ur=...` evaluates compatibility against the requested granule.
- [ ] `/compatibility` responses include the sampled `granule_ur`.
- [ ] Existing `/compatibility` requests without `granule_ur` behave as they do today.
- [ ] Tests cover forwarding, use, and response serialization of the optional parameter.

hrodmn:
This makes it possible to test the compatibility of a specific granule, which can be useful since we allow users to make tile requests for specific granules.

I also tucked in a change to one of the pydantic models for CMR collection search results to solve a model validation error when one of the fields was missing.

broken request: `https://v4jec6i5c0.execute-api.us-west-2.amazonaws.com/compatibility?collection_concept_id=C1426396538-LANCEMODIS`

result:
```
{"detail":"1 validation error for CollectionSearchResponse\nitems.0.umm.SpatialExtent.HorizontalSpatialDomain.ResolutionAndCoordinateSystem.HorizontalDataResolution.GenericResolutions.0.XDimension\n  Field required [type=missing, input_value={'YDimension': 250, 'Unit': 'Not provided'}, input_type=dict]\n    For further information visit https://errors.pydantic.dev/2.13/v/missing"}
```

resolves #204

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ See also the [API documentation](https://staging.openveda.cloud/api/titiler-cmr/
 
 - Render tiles from assets discovered via queries to [NASA's CMR](https://cmr.earthdata.nasa.gov/search)
 - Two backends: **xarray** (`/xarray`) for NetCDF/HDF5 datasets and **rasterio** (`/rasterio`) for GeoTIFF/COG assets
-- `/compatibility` helps inspect sample granules, including optional `granule_ur` and `group` selection plus lightweight group discovery for hierarchical HDF5 and NetCDF assets
+- `/compatibility` helps inspect sample granules, including optional `granule_ur`, `group`, and `skip_variable_statistics` controls plus lightweight group discovery for hierarchical HDF5 and NetCDF assets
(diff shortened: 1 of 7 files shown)


chuckwondo (approved):
Nice additions!

PR #211, opened 2026-07-02
chore: update compatibility report

hrodmn:
Good news! We are up to 931 compatible collections (from 719 from the last run).

This pairs with the output generated by the titiler-cmr-compatibility repo (https://github.com/developmentseed/titiler-cmr-compatibility/pull/2)

I built the docs site with the new report: https://developmentseed.org/titiler-cmr/chore-update-compatibility-report/compatibility/report/

resolves https://github.com/NASA-IMPACT/veda-odd/issues/397


chuckwondo (approved):
Nice! LGTM!

abarciauskas-bgse:
The report LGTM, but I have a few questions/observations about the results I want to jot down before I forget:

1. In the original report there were 1208 datasets found to be incompatible because of "group structure" but the re-run found 212 more datasets to be compatible. Recognizing that we probably can't claim for certain, without an evaluation of the parquet files from both reports, that all 212 newly compatible dataset were all from the original report's group structure classification, do you have a sense of if the failures for datasets with group structure were due to "missing spatial coordinate metadata"?
2. I wonder if it might be worth doing a straight comparison of the parquet files to make sure that none of the datasets that were deemed incompatible in this report were found compatible during the first report (and vice versa).
3. I wasn't familiar with what an IMDS token is, might be worth including an explanation.

PR #215, opened 2026-08-26
docs: add categorical statistics examples

### LINKED ISSUE #213: Expand the statistics endpoint documentation
Before we expand the usage of titiler-cmr's statistics endpoint it would be great to expand the documentation site to include an example of how to use it for categorical datasets: `categorical=True`. A good example dataset for this will be something like the Opera Disturbance collection (`collection_concept_id=C2519119034-LPCLOUD`).

Another good thing to test/explore is the `max_size` query parameter that controls the maximum size of the array that gets loaded by rio-tiler. We should have a sense for what the default behavior is and how it can be used to control the memory footprint of the statistics calculation in the Lambda function.

chuckwondo:
Covers the rasterio backend and asset selection, categorical statistics, the OPERA DIST-ANN class codes, aligning timesteps with c, and the parameters that change the numbers: max_size, reproject, and dst_crs.

Closes #213


chuckwondo:
@hrodmn, I found the missing cells that I mentioned on our call today. Turns out that they hadn't gone missing. Everything was still there, but a bug in VS Code's notebook rendering caused them to disappear from view.  A bit of digging revealed that a missing heading in the first column of a table was the problem.  Rendering in jupyter lab worked fine to begin with, but VS Code required me to add a heading to make the table show up.  Ugh!

Anyway, I force pushed an amended commit that fixes the rendering problem in VS Code.

hrodmn (approved):
Thanks for this addition @chuckwondo! Like we discussed yesterday I think we could bring your breakdown of the `categorical` + `c` + `max_size` + `dst_crs` args to the upstream docs in the titiler repo cc @vincentsarago

